PR #3, opened 2017-12-07
Features/megatron

la3lma:
More work on reducing technical debt by fixing issues discovered by static code analysis.


la3lma:
Code in a somewhat better state now, much yet to be done.

PR #10, opened 2017-12-11
Features/projectlombok

la3lma:
Let's see how much we can simplify code/make it easier to read/simpler to test by trying out some of the simpler features of projectlombok.


la3lma:
Eliminate a bunch of boilerpalte. It's all good.

PR #12, opened 2018-01-09
Moved the prime module to a separate sub-module

vihangpatil:
Moved the prime module to a separate sub-module under the ostelco-core parent module


la3lma:
confirmed

PR #27, opened 2018-03-05
Increase Credit-Control-Request support

mpeterss:
This will add more information from the Credit control requests.

The ocs-api has been updated with the following changes.

* Requests has been renamed to CreditControlRequest and the reply to CreditControlAnswer to match the DIAMETER part.

* CreditControlRequest now include both the MSISDN and IMSI

*  CreditControlRequest include Service-Information like SGSN-MNC-MCC and APN (Called-Station-Id)

* Both CreditControlRequest and CreditControlAnswer support Multi-Service-Credit-Control with requested, used and granted bytes. There is also support for Service-Id and Rating-Group.

* CreditControlAnswer now suport Final-Unit-Indication.

* First cut at support for Final-Unit-Action restrict and redirect.

Currently there is only support for one Multi-Service-Credit-Control. So the first one in the list will be used to make reservations and update usage.

### DIFF
--- a/prime/src/main/java/com/telenordigital/prime/disruptor/PrimeEvent.java
+++ b/prime/src/main/java/com/telenordigital/prime/disruptor/PrimeEvent.java
@@ -17,13 +17,22 @@ public final class PrimeEvent {
     private String msisdn;
 
     /**
-     *  Origin of word 'bucket' - PGw consumes data in `buckets` of 10 MB ~ 100 MB at a time
-     *  This field is used in...
-     *  1. Request to fetch bucket
-     *  2. Response of actual allowed bucket
-     *  3. Request to return unused bucket
+     *  Origin of word 'bucket' - P-GW consumes data in `buckets` of 10 MB ~ 100 MB at a time
+     *  This field is used in.
+     *  Request to reserve a new bucket of bytes
      */
-    private long bucketBytes;
+    private long requestedBucketBytes;
+
+    /**
+     *  Bytes that has been used from the bucket (previously reserved).
+     */
+    private long usedBucketBytes;
+
+
+    /**
+     *  Buckets that has been reserved from the bundle.
+     */
+    private long reservedBucketBytes;
 
     /**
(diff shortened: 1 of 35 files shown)


vihangpatil:
This is one major technical upgrade from the current setup. It will result in a more accurate accounting of the data consumption.

prime/src/main/java/com/telenordigital/prime/disruptor/PrimeEvent.java
```diff
@@ -17,13 +17,22 @@
...
-     *  This field is used in...
-     *  1. Request to fetch bucket
-     *  2. Response of actual allowed bucket
-     *  3. Request to return unused bucket
+     *  Origin of word 'bucket' - P-GW consumes data in `buckets` of 10 MB ~ 100 MB at a time
+     *  This field is used in.
+     *  Request to reserve a new bucket of bytes
      */
-    private long bucketBytes;
+    private long requestedBucketBytes;
```
vihangpatil: Originally, the PrimeEvent was a generic event handled by Prime. Now, it is heavily inclined towards being a PGW request and not much of Topup or Get Balance. Current changes would not break anything. But we should refactor this in future.

prime/src/main/java/com/telenordigital/prime/ocs/OcsState.java
```diff
@@ -11,12 +11,14 @@
...
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.telenordigital.prime.disruptor.PrimeEventMessageType.*;
 
 public final class OcsState implements EventHandler<PrimeEvent> {
 
     private static final Logger LOG = LoggerFactory.getLogger(OcsState.class);
 
     private final Map<String, Long> dataPackMap = new HashMap<>();
+    private final Map<String, Long> bucketReservedMap = new HashMap<>();
```
vihangpatil: This also needs evolution in future. The ocs-api will now support different types of bundle consumptions, but the in-memory bundles are still fixed to just one type - DATA.

mpeterss:
This is now merged into diameter-stack branch

PR #29, opened 2018-03-12
Diameter stack

vihangpatil:


mpeterss:
Build failed : 
    > Could not find org.jetbrains.kotlin:kotlin-stdlib-jdk8:.

mpeterss:
This is now a PR for the develop branch. Maybe we should first bring it into feature/grpc-update first as that is what it is a branch from?

PR #31, opened 2018-03-14
Proxy DataSource

mpeterss:


ocsgw/src/main/java/com/telenordigital/ocsgw/OcsServer.java
```diff
@@ -40,15 +41,21 @@ public void init(Stack stack, AppConfig appConfig) {
...
-                logger.info("Using GrpcDataSource");
+                LOG.info("Using GrpcDataSource");
                 source = new GrpcDataSource(appConfig.getGrpcServer(), appConfig.encryptGrpc());
                 break;
             case DataSourceType.LOCAL:
-                logger.info("Using LocalDataSource");
+                LOG.info("Using LocalDataSource");
                 source = new LocalDataSource();
                 break;
+            case DataSourceType.PROXY:
```
vihangpatil: As of now, PROXY is just a proxy for one type - GRPC.
In future, if we add new type - PUBSUB, then it should be configurable to choose secondary source for PROXY.
But, as of now, this is not worth.

ocsgw/src/main/java/com/telenordigital/ocsgw/data/grpc/GrpcDataSource.java
```diff
@@ -183,10 +181,22 @@ private CreditControlAnswer createCreditControlAnswer(CreditControlAnswerInfo re
...
             multipleServiceCreditControls.add(convertMSCC(mscc));
+            updateBlockedList(mscc, response.getMsisdn());
         }
         return new CreditControlAnswer(multipleServiceCreditControls);
     }
 
+    private void updateBlockedList(com.telenordigital.prime.ocs.MultipleServiceCreditControl msccGRPC, String msisdn) {
+        // This suffers from the fact that one Credit-Control-Request can have multiple MSCC
+        if (msccGRPC != null && msisdn != null) {
+            if (msccGRPC.getGranted().getTotalOctets() == 0) {
```
vihangpatil: Not only zero. Even when the granted is greater than zero, it MAY be blocked.
A better condition would be if "granted < requested".
This would imply that this might be the last bucket.

mpeterss:
https://github.com/ostelco/ostelco-core/blob/develop/prime/src/main/java/com/telenordigital/prime/ocs/EventHandlerImpl.java#L76

It is not set

Issue #59, opened 2018-04-09
[Prime] Does not handle Terminate correctly


mpeterss:
There is no return from Prime to OCSGW when the P-GW sends Credit-Control-Request Terminate.
Need to verify if this is actually correct.

ocsgw          | 09 Apr 2018 11:46:15,924 INFO  GrpcDataSource - [>>] Requesting bytes for 4790300157
ocsgw          | 09 Apr 2018 11:46:15,926 DEBUG NettyClientHandler - [id: 0x7e647cc8, L:/172.16.238.1:34767 - R:172.16.238.4/172.16.238.4:8082] OUTBOUND DATA: streamId=3 padding=0 endStream=false length=95 bytes=000000005a08031232303030332d736573736d67722e47795f4469616d657465723b31303032303030313b3235393b35616362346136662d3330321a0a343739...
prime          | INFO  [2018-04-09 11:46:15,934] org.ostelco.prime.ocs.OcsGRPCService: Received Credit-Control-Request request :: for MSISDN: 4790300157 with request id: 0003-sessmgr.Gy_Diameter;10020001;259;5acb4a6f-302
prime          | ERROR [2018-04-09 11:46:15,935] org.ostelco.prime.disruptor.PrimeEventProducer: Received empty list

mpeterss:
This is when user put phone in Flightmode

mpeterss:
ocsgw          | 09 Apr 2018 11:46:15,878 INFO  DiameterUtilities - 263 : Session-Id (UTF8String) . . . . . . . . . . 0003-sessmgr.Gy_Diameter;10020001;259;5acb4a6f-302
ocsgw          | 09 Apr 2018 11:46:15,880 INFO  DiameterUtilities - 264 : Origin-Host (OctetString) . . . . . . . . . 0003-sessmgr.Gy_Diameter
ocsgw          | 09 Apr 2018 11:46:15,880 INFO  DiameterUtilities - 296 : Origin-Realm (OctetString). . . . . . . . . loltel
ocsgw          | 09 Apr 2018 11:46:15,881 INFO  DiameterUtilities - 283 : Destination-Realm (OctetString) . . . . . . loltel
ocsgw          | 09 Apr 2018 11:46:15,881 INFO  DiameterUtilities - 258 : Auth-Application-Id (Unsigned32). . . . . . 4
[...]

Issue #56, opened 2018-04-09
[Prime][OCSgw]CCR graceful shutdown does not seem to work correctly


mpeterss:
Currently I send the following Credit-Control-Answer when the user has no more credit:

09 Apr 2018 10:34:16,946 INFO  CreditControlContext - Credit-Control-Answer
09 Apr 2018 10:34:16,946 INFO  DiameterUtilities - 263 : Session-Id (UTF8String) . . . . . . . . . . 0004-sessmgr.Gy_Diameter;15020002;516;5acb41a8-402
09 Apr 2018 10:34:16,946 INFO  DiameterUtilities - 258 : Auth-Application-Id (Unsigned32). . . . . . 4
09 Apr 2018 10:34:16,946 INFO  DiameterUtilities - 268 : Result-Code (Unsigned32). . . . . . . . . . 2001
09 Apr 2018 10:34:16,947 INFO  DiameterUtilities - 416 : CC-Request-Type (Integer32) . . . . . . . . 1
09 Apr 2018 10:34:16,947 INFO  DiameterUtilities - 415 : CC-Request-Number (Unsigned32). . . . . . . 0
09 Apr 2018 10:34:16,948 INFO  DiameterUtilities - 264 : Origin-Host (OctetString) . . . . . . . . . 0004-sessmgr.Gy_Diameter
09 Apr 2018 10:34:16,948 INFO  DiameterUtilities - 296 : Origin-Realm (OctetString). . . . . . . . . loltel
09 Apr 2018 10:34:16,949 INFO  DiameterUtilities - 456 : Multiple-Services-Credit-Control (Grouped). <Grouped>
09 Apr 2018 10:34:16,949 INFO  DiameterUtilities -   432 : Rating-Group (Unsigned32) . . . . . . . . 10
09 Apr 2018 10:34:16,949 INFO  DiameterUtilities -   439 : Service-Identifier (Unsigned32) . . . . . 1
09 Apr 2018 10:34:16,949 INFO  DiameterUtilities -   431 : Granted-Service-Unit (Grouped). . . . . . <Grouped>
09 Apr 2018 10:34:16,949 INFO  DiameterUtilities -     412 : CC-Input-Octets (Unsigned64). . . . . . 0
09 Apr 2018 10:34:16,952 INFO  DiameterUtilities -     414 : CC-Output-Octets (Unsigned64) . . . . . 0
[...]

PR #61, opened 2018-04-11
Feature/ccr terminate

mpeterss:
This is for issue : 
    https://github.com/ostelco/ostelco-core/issues/59
    https://github.com/ostelco/ostelco-core/issues/56


mpeterss:
So far this seems to solve issue https://github.com/ostelco/ostelco-core/issues/59 but not #56

PR #80, opened 2018-04-26
Fix/topup on startup

mpeterss:


mpeterss:
Fix is the first commit. The second it just to remove the naming confusion.

PR #99, opened 2018-05-08
Cloud function for push notifications on low balance

mpeterss:
When the user is low on balance the server (Prime) will add to firebase path /fcm-requests/{msisdn}
{
    msisdn : 4753945
    noOfBytes: 545789345
}

This will be picked up by a Firebase cloud function that will send a push notification.

The application will put its firebase token in the path: /firebaseCloudMessage/
{
    msisdn: token
}


mpeterss:
Tested with my Android test app and it does work :)

mpeterss:
For Android test client: https://github.com/ostelco/ostelco-android

mpeterss:
Ref for test client : https://github.com/ostelco/ostelco-android

la3lma:
In general, this looks very interesting. I have a few general comments though:

Is there a rule to regulate access to the path in firebase that is used for 

-  /fcm-test/{msisdn} is imho not a good name :-)   It should be more descriptive.
- I couldn't see any security rules protecting /fmc-test or /firebaseCloudMessage/ ? Is that relevant?
- I would like to see documentation that includes a sequence diagram. I'd be perfectly willing to sit down and make one with you :-)

mpeterss:
path was actually 'fcm-requests' but description still was fcm-test :)

mpeterss:
Fore security, what we could do is that the user can only store its FirebaseToken in /firebaseCloudMessage/{msisdn} that match his own msisdn.

For sending a notification. That is adding a new item in /fcm-requests/{msisdn} that is done by Prime, so this would have to be restricted by admin token access

PR #104, opened 2018-05-11
Guard agains sending CCA more then once for one CCR

mpeterss:


mpeterss:
Will think a bit more about this, it seems a bit to hacky

PR #106, opened 2018-05-14
Updates to client API as agreed upon in discussions (WIP) 

kmmatwork:


client-api/src/main/java/org/ostelco/topup/api/db/SubscriberDAOImpl.java
```diff
@@ -50,17 +49,17 @@
...
     @Override
-    public Either<Error, List<Offer>> getOffers(final String subscriptionId) {
+    public Either<Error, List<Product>> getProducts(final String subscriptionId) {
         return Either.left(new Error());
     }
 
     @Override
-    public Option<Error> acceptOffer(final String subscriptionId, final String offerId) {
+    public Option<Error> acceptProduct(final String subscriptionId, final String productId) {
         return Option.of(null);
```
vihangpatil: Maybe `purchaseProduct` sounds better than `acceptProduct`.

client-api/src/main/java/org/ostelco/topup/api/db/SubscriberDAOImpl.java
```diff
@@ -50,17 +49,17 @@
...
     }
 
     @Override
-    public Option<Error> rejectOffer(final String subscriptionId, final String offerId) {
+    public Option<Error> rejectProduct(final String subscriptionId, final String productId) {
         return Option.of(null);
```
vihangpatil: We might not need `rejectProduct` functionality.

client-api/src/main/java/org/ostelco/topup/api/resources/ProductsResource.java
```diff
@@ -24,50 +24,50 @@
...
     @PUT
-    @Path("{offer-id}")
+    @Path("{product-id}")
     @Produces({"application/json"})
-    public Response updateOffer(@Auth AccessTokenPrincipal token,
+    public Response updateProduct(@Auth AccessTokenPrincipal token,
             @NotNull
-            @PathParam("offer-id") String offerId,
+            @PathParam("product-id") String productId,
             @DefaultValue("true") @QueryParam("accepted") boolean accepted) {
```
vihangpatil: I think we should have `purchaseProduct` as `POST` method and without `boolean accepted`.
Also, actual API might be much more complex than this since it needs integration with Payment Gateway.

kmmatwork:
All of this makes sense. Updated.

PR #111, opened 2018-05-15
Adds an ugly 'in-memory' backend DB for test of 'profile' API

kmmatwork:


client-api/src/main/java/org/ostelco/topup/api/core/Profile.java
```diff
@@ -5,9 +5,15 @@
 
 @Data
 public class Profile {
```
la3lma: Class should always have a short, concise class comment that describes the intent of the class

client-api/src/main/java/org/ostelco/topup/api/db/SubscriberDAOInMemoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.NonNull;
+
+/**
+ * An 'in-memory' store for testing.
+ *
+ */
+public class SubscriberDAOInMemoryImpl implements SubscriberDAO {
```
la3lma: final?

client-api/src/main/java/org/ostelco/topup/api/db/SubscriberDAOInMemoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /* Table for 'profiles'. */
+    private final Map<String, Profile> profileTable = new HashMap<>();
```
la3lma: Actually, this needs to be thread safe.  The requests coming into the storage may produced by multiple threads, and may overlap in time.  I suggest either using a concurrent datastruture (e.g. https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html implemented in https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html), or to put a lock around  the accesses to the profileTable.   A simple "spin lock" (e.g. an Object instance used in a "synchronised" block) can easily lead to lock contention, but is is simple to implement, so if you can't easily fit in a concurrent datastructure I suggest using a spin lock.   For production use, if  a concurrent datastructure isn't usable, I would recommend using read/write locks, that can be quite efficient if done right.    The in-memory thing won't be used in production so that isn't very important, but avoiding race conditions even in test is .-)

Issue #102, opened 2018-05-11
[OCSgw] [Prime] Reconnect gRPC 


mpeterss:
It seems like we are loosing the gRPC connection between OCSgw and Prime sometimes and when that happen it is not re established again until ocsgw is restarted.

Last item in logs seems to be an error on the streamobserver: 

     https://github.com/ostelco/ostelco-core/blob/develop/ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java#L85

We also have a dependency that prime needs to be started before ocsgw.

mpeterss:
This seems to work fine now :)

PR #113, opened 2018-05-16
Reconnect gRPC on error

mpeterss:


mpeterss:
https://github.com/ostelco/ostelco-core/issues/102

mpeterss:
>Should we call reconnect on any specific type of Throwable/Exception, or should it be called for any type?

I only see "io.grpc.StatusRuntimeException: UNAVAILABLE: io exception", but maybe I should start with only that one then

> If the onError event occurs how do we guarantee that messages are not lost, and they are sent again after reconnection?

Currently all requests are lost when gRPC is down, nothing is buffered. If we buffer, we need to figure out how we long should store a request before we throw it away.

ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java
```diff
@@ -77,9 +81,22 @@ protected boolean removeEldestEntry(Map.Entry<String, SessionContext> eldest) {
...
-    private abstract class AbstactObserver<T> implements StreamObserver<T> {
+    private abstract class CreditControlRequestObserver<T> implements StreamObserver<T> {
+        public final void onError(Throwable t) {
+            LOG.error("We got an error", t);
+            reconnectCreditControlRequest();
+        }
+
+        public final void onCompleted() {
+            // Nothing to do here
+            LOG.info("It seems to be completed");
```
prasanthu: Looks unnecessary

ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java
```diff
@@ -77,9 +81,22 @@ protected boolean removeEldestEntry(Map.Entry<String, SessionContext> eldest) {
     };
 
 
-    private abstract class AbstactObserver<T> implements StreamObserver<T> {
+    private abstract class CreditControlRequestObserver<T> implements StreamObserver<T> {
+        public final void onError(Throwable t) {
+            LOG.error("We got an error", t);
```
prasanthu: Can you change the text to anything other than `We got an error` ;-)

ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java
mpeterss: Yes I just wanted to check if it ever got called. So far it has never been. I will remove it

PR #136, opened 2018-05-23
restart the keepalive timer when gRPC is recreated

mpeterss:


ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java
```diff
@@ -133,6 +133,8 @@ private void reconnectCreditControlRequest() {
             initCCRFuture = executorService.schedule((Callable<Object>) () -> {
                         LOG.info("Calling initCreditControlRequest");
                         initCreditControlRequest();
+                        LOG.info("Calling initKeepAlive");
+                        initKeepAlive();
```
vihangpatil: So, this is called only once?

vihangpatil: Cannot call initKeepAlive() directly. The method starts a scheduled executor. So, to call it again, need to check if the reference is not null. If not null, then shutdown the existing executor before starting again.

mpeterss: yes, it is already done :
    https://github.com/ostelco/ostelco-core/blob/develop/ocsgw/src/main/java/org/ostelco/ocsgw/data/grpc/GrpcDataSource.java#L127

PR #141, opened 2018-05-25
Features/enable tsl on grpc

la3lma:
Enable TSL for the ocs/prime interface.


la3lma:
We should merge and deploy this only after making very certain that it won't break anything, and when we do deploy, et should be early in the morning with a full crew present so that we we can back out/fix issues quickly if/when they appear.

vihangpatil:
We "MIGHT" be able to test this locally using docker / docker-compose.
I have managed to run the Google's Cloud Endpoints / ESP (Extensible Service Proxy) as docker container on my system for gRPC traffic.
If we pass the SSL/TLS certificates and secrets via mounted volume to ESP, and set DNS entry in `ocsgw` to point to `esp`, it might run locally.
Only risk/limitation I see is that ESP fetches `service spec` published in GCP. It is not possible to locally pass new "untested" version of spec without publishing it. Hopefully, there are no changes in `service spec`.

PR #152, opened 2018-06-06
Add pseudonym API  for client

prasanthu:


client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupModule.kt
```diff
@@ -37,27 +39,33 @@ class TopupModule : PrimeModule {
...
 
         val dao = SubscriberDAOImpl(storage, ocsSubscriberService)
         val jerseyEnv = env.jersey()
 
+        val client: Client = JerseyClientBuilder(env)
+                .using(config.jerseyClientConfiguration)
+                .using(ObjectMapper()
+                        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false))
+                .build(env.getName())
+        client.property(ClientProperties.CONNECT_TIMEOUT, 2000)
```
kmmatwork: Not really needed as the timeout values are pulled in from the config file through the `config.jerseyClientConfiguration` call. If values different from Dropwizards default are needed then it is just to update the config. See:
https://www.dropwizard.io/1.3.2/docs/manual/configuration.html#man-configuration-clients-http

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupConfiguration.kt
```diff
@@ -23,4 +23,6 @@ class TopupConfiguration {
     fun setAuthenticationCachePolicy(spec: String) {
         this.authenticationCachePolicy = CacheBuilderSpec.parse(spec)
     }
+    var pseudonymEndpoint = ""
```
kmmatwork: Should be in the config file I think - in `prime/config/config.yaml` to be specific. And if not set in config then the app. should fail at startup due to invalid config.

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupModule.kt
```diff
@@ -37,27 +39,33 @@ class TopupModule : PrimeModule {
...
+        client.property(ClientProperties.READ_TIMEOUT, 2000)
+
         /* APIs. */
         jerseyEnv.register(AnalyticsResource(dao))
         jerseyEnv.register(ConsentsResource(dao))
         jerseyEnv.register(ProductsResource(dao))
         jerseyEnv.register(ProfileResource(dao))
-        jerseyEnv.register(SubscriptionResource(dao))
+
+        LOG.info("client-api pseudonymEndpoint ${config.pseudonymEndpoint}")
```
kmmatwork: No need for log if `pseudonymEndpoint` is set in `prime/config/config.yaml`.

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupModule.kt
```diff
@@ -37,27 +39,33 @@ class TopupModule : PrimeModule {
...
+        jerseyEnv.register(SubscriptionResource(dao, client, config.pseudonymEndpoint))
 
         /* For reporting OAuth2 caching events. */
         val metrics = SharedMetricRegistries.getOrCreate(env.getName())
 
-        val client: Client = JerseyClientBuilder(env)
-            .using(config.jerseyClientConfiguration)
-            .using(ObjectMapper()
-                    .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false))
-            .build(env.getName())
```
kmmatwork: Double empty line.

client-api/src/main/kotlin/org/ostelco/prime/client/api/store/SubscriberDAOImpl.kt
```diff
@@ -98,6 +98,20 @@ class SubscriberDAOImpl(private val storage: Storage, private val ocsSubscriberS
 
     }
 
+    override fun getMsisdn(subscriptionId: String): Either<ApiError, String> {
+        var msisdn: String? = null
+        try {
+            msisdn = storage.getSubscription(subscriptionId)
```
kmmatwork: That `storage.getSubscription()` should return the MSISDN is not that obvious. Maybe the "storage" class should be extended with a `getMsisdn()` method?

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupConfiguration.kt
prasanthu: This is already in that `config.yaml`. Where should we test this, in `prime` or `client-api` module ?

client-api/src/main/kotlin/org/ostelco/prime/client/api/store/SubscriberDAOImpl.kt
prasanthu: currently subscription is` msisdn` ?

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupConfiguration.kt
prasanthu: @kmmatwork  What do you mean by fail, Should I thrown an exception from `TopupModule.kt` ?

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupModule.kt
prasanthu: Done

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupModule.kt
prasanthu: Done

client-api/src/main/kotlin/org/ostelco/prime/client/api/store/SubscriberDAOImpl.kt
prasanthu: Done

client-api/src/main/kotlin/org/ostelco/prime/client/api/TopupConfiguration.kt
kmmatwork: Yes, add a `@NotNull` annotation and it will stop prime form starting if not configured correctly.

Not sure about testing as we currently don't have tests for configs in the api module. But the config is simple and it might be a bit of overkill to have tests for it?

kmmatwork:
Two more concerns:

1) I'm a bit sceptical  about having large timeout values in Jersey Client, as it might hide issues. Better to use the default values and expose the issues that this might cause. Could be that the issues could be fixed in other ways than?

2) Codacy compliaints should be fixed (even if some/most of them seems stupid). Helps with keeping the code somewhat consistent.

PR #171, opened 2018-06-15
Check if tag is added to master branch for cloud builder

vihangpatil:


prime/script/check_repo.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# This script checks if current branch is master
+
+if [ master != $1 ];
+then
```
la3lma: I usually prefer the style "if [ master != $1 ] ; then"  (not breaking the line before "then")  and also to wrap the $1 in quotes "$1", just to make sure that the expansion is not treated as code.     Details, but still ... :-)

prime/script/deploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ ! -f prime/script/deploy.sh ]; then
+    echo "Run this script from project root dir (ostelco-core)"
```
la3lma: Pedantic comment, but use '(>&2 echo "Run this script ...")'.  The rationale being:
a) Error message to stderr
b) Do it in a subshell to avoid interacting with other redirects, if they are in action.

prime/script/deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+
+export PROJECT_ID="$(gcloud config get-value project -q)"
+export PRIME_VERSION="$(gradle prime:properties -q | grep "version:" | awk '{print $2}' | tr -d '[:space:]')"
+export BRANCH_NAME=$(git branch | grep \* | cut -d ' ' -f2)
+
+echo PROJECT_ID=${PROJECT_ID}
+echo PRIME_VERSION=${PRIME_VERSION}
+echo BRANCH_NAME=${BRANCH_NAME}
+
+prime/script/check_repo.sh ${BRANCH_NAME}
```
la3lma: Being pedantic, I would prefer testing the existence of this script near the top  of the file and bailing out if it doesn't exist. Also I'd put it in a variable and say something like

   $CHECK_REPO ${BRANCH_NAME}

... but again, this is just me being pedantic.

PR #180, opened 2018-06-26
ocsgw logs uploaded to gcp stackdriver

vihangpatil:


ocsgw/script/start.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-# Start app
+# Start app for production
 exec java \
     -Dfile.encoding=UTF-8 \
+    -Dlogback.configurationFile=/config/logback.xml \
```
mpeterss: Also needs to be added to pack task in gradle

mpeterss:
LGTM

PR #217, opened 2018-08-16
Feature/cleanup payment

prasanthu:


payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -168,9 +154,6 @@ class StripePaymentProcessor : PaymentProcessor {
                 Customer.retrieve(customerId).sources.retrieve(sourceId).delete().id
             }
 
-    private fun isSourceStored(customerId: String, sourceId: String): Either<ApiError, Boolean> =
```
mpeterss: Might be we need this again incase we need to check if a source is already stored before adding it in the charge function. That is where it was used before, but this did not work with tokens as they are converted to a source after they have been added.

prasanthu: We can bring it back when needed

PR #218, opened 2018-08-17
Feature/stripe testing

mpeterss:
Added integration tests for the StripePaymentProcessor.

As this needs a Strip API Key it will only be run from developer machines and not by Travis


payment-processor/build.gradle
```diff
@@ -1,6 +1,20 @@
 plugins {
   id "org.jetbrains.kotlin.jvm" version "1.2.50"
   id "java-library"
+  id "idea"
```
prasanthu: Do we need idea plugin ?

mpeterss: Not needed, but nice to be able to run the test in IntelliJ

PR #238, opened 2018-08-30
Renamed analytics to dataflow-pipelines.

vihangpatil:


dataflow-pipelines/src/main/kotlin/org/ostelco/dataflow/pipelines/DeployPipeline.kt
```diff
@@ -0,0 +1,63 @@
...
+fun main(args: Array<String>) {
+    System.setProperty(ContextInitializer.CONFIG_FILE_PROPERTY, "config/logback.xml")
+    DeployPipeline().deploy(pipelineName = "DATA_CONSUMPTION")
+}
+
+class DeployPipeline {
+
+    private fun parseOptions(): PipelineOptions {
+
+        // may be we need to pass options via command-line args
```
la3lma: I really dislike dead code:-)

dataflow-pipelines/src/main/kotlin/org/ostelco/dataflow/pipelines/DeployPipeline.kt
```diff
@@ -0,0 +1,63 @@
...
+enum class PipelineDefinitionRegistry(val pipelineDefinition: PipelineDefinition) {
+    DATA_CONSUMPTION(DataConsumptionPipelineDefinition),
+    DUMMY(DummyPipelineDefinition),
+}
+
+fun main(args: Array<String>) {
+    System.setProperty(ContextInitializer.CONFIG_FILE_PROPERTY, "config/logback.xml")
+    DeployPipeline().deploy(pipelineName = "DATA_CONSUMPTION")
+}
+
```
la3lma: Near the top of this class, because it is an important one,  there should be a descriptive text describing at least the intent behind the class and possibly (more important in other cases) some examples of happy day use cases.  For this class it may be of interest to list the command line parameters that are overridden (that would make the dead code below less necessary to understand the usage of the class).

dataflow-pipelines/src/main/kotlin/org/ostelco/dataflow/pipelines/DeployPipeline.kt
```diff
@@ -0,0 +1,63 @@
...
+    val options = PipelineOptionsFactory
+            .fromArgs(
+                    "--project=pantel-2decb",
+                    "--runner=DataflowRunner",
+                    "--stagingLocation=gs://data-traffic/staging/",
+                    "--jobName=data-traffic")
+            .withValidation()
+            .create()
+    */
+
```
la3lma: just say "// Ignore command line parameters, set options explicitly instead."

dataflow-pipelines/src/main/kotlin/org/ostelco/dataflow/pipelines/definitions/PipelineDefinition.kt
```diff
@@ -0,0 +1,7 @@
+package org.ostelco.dataflow.pipelines.definitions
+
+import org.apache.beam.sdk.Pipeline
+
+interface PipelineDefinition {
+    fun define(pipeline: Pipeline)
+}
```
la3lma: missing newline at end of file (not the only place)

PR #256, opened 2018-09-07
Adds test for 'set default source' to payment acceptance tests

kmmatwork:


acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -43,6 +43,25 @@ object StripePayment {
...
+        val customer = Customer.retrieve(customerId)
+        return customer.defaultSource
+    }
+
+    fun getCustomerIdForEmail(email: String) : String {
+
+        // https://stripe.com/docs/api/java#create_card_token
+        Stripe.apiKey = System.getenv("STRIPE_API_KEY")
+
+        val customers = Customer.list(emptyMap()).data
```
mpeterss: We are currently not storing the email with the customer id in Stripe. So how do this work ?

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -43,6 +43,25 @@ object StripePayment {
         return token.card.id
     }
 
+    fun getDefaultSourceForCustomer(customerId: String) : String {
```
mpeterss: Why did we reimplement the functions that we have created APIs for?

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -23,24 +23,32 @@ object StripePayment {
         return token.id
     }
 
-    fun createPaymentSourceId(): String {
+    fun getCardIdForTokenId(tokenId: String) : String {
```
mpeterss: I think this function is already implemented in our API

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -23,24 +23,32 @@ object StripePayment {
...
-        // TODO martin: set valid map values
-        val sourceMap = mapOf<String,Any>()
-        val source = Source.create(sourceMap)
-        return source.id
+        val token = Token.retrieve(tokenId)
+        return token.card.id
     }
 
-    fun getCardIdForTokenId(tokenId: String) : String {
+    fun getDefaultSourceForCustomer(customerId: String) : String {
```
mpeterss: I think this function is already implemented in our API or do we only have setDefaultSource ?

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -256,6 +257,94 @@ class SourceTest {
...
+    @Test
+    fun `okhttp test - GET list sources`() {
+
+        StripePayment.deleteAllCustomers()
+        Firebase.deleteAllPaymentCustomers()
+
+        val email = "purchase-${randomInt()}@test.com"
+        createProfile(name = "Test Payment Source", email = email)
+
+        val sourceId = StripePayment.createPaymentTokenId()
```
mpeterss: rename to tokenId

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -256,6 +257,94 @@ class SourceTest {
...
+        // Ties source with user profile both local and with Stripe
+        post<PaymentSource> {
+            path = "/paymentSources"
+            subscriberId = email
+            queryParams = mapOf("sourceId" to sourceId)
+        }
+
+        Thread.sleep(200)
+
+        val newSourceId = StripePayment.createPaymentTokenId()
```
mpeterss: val newTokenId

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -256,6 +257,94 @@ class SourceTest {
...
+
+        // TODO: Update to fetch the Stripe customerId from 'admin' API when ready.
+        val customerId = StripePayment.getCustomerIdForEmail(email)
+
+        // Verify that original 'sourceId/card' is default.
+        assertEquals(cardId, StripePayment.getDefaultSourceForCustomer(customerId),
+                "Expected $cardId to be default source for $customerId")
+
+        // Set new default card.
+        put<PaymentSource> {
```
mpeterss: Is the API for setting a source as Default the same as for adding a source?

mpeterss: nm, it is a post to add new. Put to set default :)

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
mpeterss: ok, I see we have implementation in org.ostelco.prime.paymentprocessor but not exposed through client API

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
kmmatwork: It is used to check whether the "setDefaultSource" API actually works as intended (by obtaining the information directly from Stripe). But bad naming perhaps?

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
kmmatwork: Good point! But this is also inconsistent in the API itself I think. Should perhaps be cleaned up?

PR #267, opened 2018-09-11
Fix/payment api error code

mpeterss:
Adding PaymentApi instead of using ApiError everywhere. This is the first part of refactoring our error classes to use our own error codes.


prime-api/src/main/kotlin/org/ostelco/prime/paymentprocessor/core/PaymentError.kt
```diff
@@ -0,0 +1,19 @@
+package org.ostelco.prime.paymentprocessor.core
+
+import javax.ws.rs.core.Response
+
+sealed class PaymentError(val description: String) {
+    open var status : Int = 0
+}
+
+class ForbiddenError(description: String) : PaymentError(description) {
+    override var status : Int = Response.Status.NOT_FOUND.getStatusCode()
```
kmmatwork: Maybe `Response.Status.FORBIDDEN`?

prime-api/src/main/kotlin/org/ostelco/prime/paymentprocessor/core/PaymentError.kt
```diff
@@ -0,0 +1,19 @@
...
+}
+
+class NotFoundError(description: String) : PaymentError(description) {
+    override var status : Int = Response.Status.NOT_FOUND.getStatusCode()
+}
+
+class BadGatewayError(description: String) : PaymentError(description) {
+    override var status : Int = Response.Status.BAD_REQUEST.getStatusCode()
+}
```
kmmatwork: Missing line shift...

prime-api/src/main/kotlin/org/ostelco/prime/paymentprocessor/core/PaymentError.kt
mpeterss: Thanks. I will fix

PR #278, opened 2018-09-12
Added errorCode to ApiError

mpeterss:
This task was to add error codes to all respons messages for the API in case the request failed.

I am not sure how we should do this. But I started something here that I would like to get some feedback on. This adds error codes to all ApiErrors. Right now the error codes are way to specific to make sense for the client. But I am also not sure why HTTP error codes are not enough.

There was also talk about the server returning error messages to the client that could be shown to the end user.  i18n on the backend etc. I have not done any work on this and I am not sure if I should. I assume it would be enough for the client to map these errorCodes to fail messages?

I was also thinking about what to do about 3rd party errors. Should I just append the Stripe error and pass it on to the client? It could possibly contain secrets..


mpeterss:
I am not sure how we should do this. But I started something here that I would like to get some feedback on. This adds error codes to all ApiErrors. 

Possibly the error messages sent to the client should also be moved as we have had requests for long and short messages. i18n on the backend etc.

mpeterss:
I am reading up on the subject:
https://nordicapis.com/best-practices-api-error-handling/

PR #284, opened 2018-09-14
Additional metrics to BQ extractor

kmmatwork:


bq-metrics-extractor/config/config.yaml
```diff
@@ -45,3 +45,21 @@ bqmetrics:
...
+  - type: gauge
+    name: active_users_right_now
+    help: Number of active users rigth now (with 1 minute interval)
+    resultColumn: count
+    sql: >
+        SELECT count(DISTINCT msisdn) AS count FROM `pantel-2decb.data_consumption.raw_consumption`
+        WHERE timestamp >= TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 1 MINUTE)
+  - type: gauge
+    name: average_data_consumed
+    help: Average data consumed in last 24 hours
```
prasanthu: We are already pushing the values required to make this metric. Isn't this, total_data_used / sims_who_have_used_data ?

PR #293, opened 2018-09-18
Store customerId in Stripe

mpeterss:
This remove the storage of the paymentId in Firebase. Instead we
will get the Id from Stripe. Each time we create a new user in
Stripe we store the customer email with Stripe. So this is used
to fetch the paymentId once needed.


payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -59,6 +62,20 @@ class StripePaymentProcessor : PaymentProcessor {
...
 
+    override fun getPaymentProfile(userEmail: String): Either<PaymentError, ProfileInfo> {
+            val customerParams = mapOf(
+                    "limit" to "1",
+                    "email" to userEmail)
+            val customerList = Customer.list(customerParams)
+            if (customerList.data.isEmpty()) {
+                 return Either.left(NotFoundError("Could not find a payment profile for user  $userEmail"))
+            } else if (customerList.data.size > 1){
+                 return Either.left(NotFoundError("Multiple profiles for user $userEmail found"))
```
prasanthu: Is this allowed in Stripe (having multiple profiles with same email ) ?

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -1,7 +1,6 @@
 package org.ostelco.at.jersey
 
 import org.junit.Test
-import org.ostelco.at.common.Firebase
```
prasanthu: Did you remove the Firebase dependency from gradle settings

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
mpeterss: I am not sure. The Documentation does not mention if it must be unique.
https://stripe.com/docs/api#create_customer
So I added a check incase :)

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
mpeterss: Thanks, I will fix this

PR #297, opened 2018-09-19
Rename prime-api to prime-modules

mpeterss:


mpeterss:
Should it be `prime-modules` or `prime-module-interfaces` ?

PR #301, opened 2018-09-20
Adds support for Stripe 'source' payment method

kmmatwork:


acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -29,7 +29,20 @@ object StripePayment {
         Stripe.apiKey = System.getenv("STRIPE_API_KEY")
 
         // TODO martin: set valid map values
```
mpeterss: Todo can be removed then :)

PR #307, opened 2018-09-20
Adds support for default segment per country (WIP)

kmmatwork:


neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -155,6 +155,7 @@ object Neo4jStoreSingleton : GraphStore {
         val bundleId = subscriber.id
 
         val either = subscriberStore.create(subscriber, transaction)
+                .flatMap { subscriberToSegmentStore.create(subscriber.id, subscriber.country.toLowerCase(), transaction) }
```
vihangpatil: Now, when it fails, it will give *Segment does not exists* error.  Need to do a `mapLeft` on `subscriberToSegmentStore.create` to change the error object/message.  Don't apply `mapLeft` after `flatMap`, since it will overwrite error from `subscriberStore.create` too.

PR #330, opened 2018-09-27
Adds support for removing payment sources

kmmatwork:


prime/infra/dev/prime-client-api.yaml
```diff
@@ -151,6 +151,28 @@ paths:
           description: "User not found."
       security:
       - auth0_jwt: []
```
vihangpatil: Also copy same changes to prime-client-api.yaml in new-dev and new-prod.

kmmatwork: Done! Merge with latest develop too and should build and pass tests now...

client-api/src/main/kotlin/org/ostelco/prime/client/api/store/SubscriberDAOImpl.kt
```diff
@@ -350,4 +350,19 @@ class SubscriberDAOImpl(private val storage: ClientDataSource, private val ocsSu
                             .mapLeft { mapPaymentErrorToApiError("Failed to list sources", ApiErrorCode.FAILED_TO_FETCH_PAYMENT_SOURCES_LIST, it) }
                 }
     }
+
+    override fun removeSource(subscriberId: String, sourceId: String): Either<ApiError, SourceInfo> {
+        return paymentProcessor.getPaymentProfile(subscriberId)
+                .fold(
+                        {
+                            paymentProcessor.createPaymentProfile(subscriberId)
```
mpeterss: if there is no payment profile then we can not delete the source. So we would not need to create it?

vihangpatil: I will change the logic so that we delete the source only if payment profile exists.  When the payment profile does not exist, should I return success or fail? @mpeterss

mpeterss: This seems like a client error so something on 4xx I think.
Maybe 400, 404 or 410 ?

PR #367, opened 2018-10-09
Add mobicents pom file manually to fix the jdk11 issue.

prasanthu:
https://stackoverflow.com/questions/52414255/maven-with-jdk-11-fails-when-downloading-a-wrong-dependency-with-extension-10

This will be fixed in java release 11.0.2


vihangpatil:
In order to avoid adding JAR files for external libraries into our repo, I created a Google Cloud Storage public bucket to publicly host the pom.xml files and the jar file from this PR.  Then I found that the simpler, even though not a perfect, fix is just to use `http` without SSL.  I have changed Travis config to build with openjdk11 and Gradle 5 (RC1) and it worked.  Closing this PR.

PR #371, opened 2018-10-10
Feature/add houston api

prasanthu:


vihangpatil:
Since this is independent application and not using any `test / library / protobuf / grpc / swagger spec / deployment` with others, it may be moved to its own repo.  That way, it can also be built via Travis.

prasanthu:
Move to another pull request https://github.com/ostelco/ostelco-core/pull/394

PR #372, opened 2018-10-11
Updates 'active users' metrics extracted from BQ

kmmatwork:
Updated/new metrics:
  - active users last 24 hours
  - active users yesterday
  - active users today

The standard Firebase events:

  - session_start : user engages with the app
  - screen_view : user switches between screens in app
  - user_engagement : periodic events when app is in foreground

are used to determine whether an user is active or not. As all
of the events are subject to some limitations (see ref. below)
all of are included in the "active user" detection.

Ref.: https://support.google.com/firebase/answer/6317485

### DIFF
--- a/bq-metrics-extractor/config/config.yaml
+++ b/bq-metrics-extractor/config/config.yaml
@@ -10,13 +10,31 @@ logging:
            level: severity
 
 bqmetrics:
-  - type: summary
-    name: active_users
-    help: Number of active users
+  - type: gauge
+    name: active_app_users_last24hours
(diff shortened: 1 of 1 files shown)


kmmatwork:
Output from test run in dev:
```
I  Querying BQ for total 22 metric values 
I  Gauge metric active_app_users_last24hours = 10 
I  Gauge metric active_app_users_yesterday = 7 
I  Gauge metric active_app_users_today = 6 
I  Gauge metric sims_who_have_used_data = 0 
I  Gauge metric total_data_used = 0 
```

PR #380, opened 2018-10-15
Active Pseudo ID for Subscriber ID

vihangpatil:


jersey/src/main/kotlin/org/ostelco/prime/jersey/PingResource.kt
```diff
@@ -0,0 +1,12 @@
+package org.ostelco.prime.jersey
+
+import javax.ws.rs.GET
+import javax.ws.rs.Path
+import javax.ws.rs.core.Response
+
+@Path("/ping")
+class PingResource {
```
prasanthu: Whats the purpose of this REST interface ?

PR #387, opened 2018-10-18
Stop zero rated products from being sent to Stripe (WIP)

kmmatwork:


payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -204,52 +204,61 @@ class StripePaymentProcessor : PaymentProcessor {
...
-        return either(errorMessage) {
-            val chargeParams = mutableMapOf(
-                    "amount" to amount,
-                    "currency" to currency,
-                    "customer" to customerId,
-                    "capture" to false)
-            if (sourceId != null) {
-                chargeParams["source"] = sourceId
+        return when {
+            amount == 0 -> Either.right("ZERO_CHARGE_IS_OK")
```
prasanthu: I think we should  make the chargeId unique (may be add a uuid at  the end e.g. ZERO_CHARGE_IS_OK_<UUID>)

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -204,52 +204,61 @@ class StripePaymentProcessor : PaymentProcessor {
...
-                    ifFalse = { ForbiddenError("Review required, $errorMessage $review") }
-            )
-        }.flatMap { charge ->
-            try {
-                charge.capture()
-                Either.right(charge.id)
-            } catch (e: Exception) {
-                logger.warn(errorMessage, e)
-                Either.left(BadGatewayError(errorMessage))
+        return when {
```
prasanthu: I think we should pass the amount parameter and check that value instead of checking the chargeId

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -204,52 +204,61 @@ class StripePaymentProcessor : PaymentProcessor {
...
+                }
             }
         }
     }
 
     override fun refundCharge(chargeId: String): Either<PaymentError, String> =
-            either("Failed to refund charge $chargeId") {
-                val refundParams = mapOf("charge" to chargeId)
-                Refund.create(refundParams).charge
+            when {
```
prasanthu: Same as the last comment, pass amount parameter here.

PR #399, opened 2018-11-14
Initial (and probably incomplete and/or buggy) implementation of the ES2+ protocol

la3lma:
Merged simply because we don't like long running branches, and also because it is not messing up anything else. Everything is confined to a single "esim" diectory.


vihangpatil:
This PR is not required.

PR #401, opened 2018-11-20
Deploy ocsgw dev on compute instance

mpeterss:


ocsgw/config/logback.xml
```diff
@@ -21,15 +21,14 @@
...
+  <logger name="org.ostelco" level="INFO"/>
+  <logger name="org.mobicents.tests.diameter" level="INFO"/>
   <logger name="org.jdiameter.common.impl.validation" level="INFO"/>
   <logger name="jdiameter.statistic" level="INFO"/>
   <logger name="org.jdiameter" level="INFO"/>
 
   <root level="INFO">
     <appender-ref ref="STDOUT"/>
-    <appender-ref ref="CLOUD_ASYNC"/>
   </root>
```
vihangpatil: Should we also remove the library/dependency which uploads logs to GCP stackdriver?

PR #404, opened 2018-11-26
Feature/houston add notification

prasanthu:


admin-api/src/main/kotlin/org/ostelco/prime/admin/api/HoustonResources.kt
```diff
@@ -209,3 +210,56 @@ class RefundsResource() {
...
+                              @NotNull
+                              @PathParam("email")
+                              email: String,
+                              @NotNull
+                              @QueryParam("title")
+                              title: String,
+                              @NotNull
+                              @QueryParam("message")
+                              message: String): Response {
+        if (token == null) {
```
mpeterss: Should we not check if the token is valid and have access rights for this API ?

admin-api/src/main/kotlin/org/ostelco/prime/admin/api/HoustonResources.kt
```diff
@@ -209,3 +210,56 @@ class RefundsResource() {
...
+
+    // TODO: Reuse the one from SubscriberDAO
+    private fun getMsisdn(subscriberId: String): Either<ApiError, String> {
+        return try {
+            storage.getMsisdn(subscriberId).mapLeft {
+                NotFoundError("Did not find msisdn for this subscription.", ApiErrorCode.FAILED_TO_STORE_APPLICATION_TOKEN, it)
+            }
+        } catch (e: Exception) {
+            logger.error("Did not find msisdn for subscriberId $subscriberId", e)
+            Either.left(BadGatewayError("Did not find subscription", ApiErrorCode.FAILED_TO_STORE_APPLICATION_TOKEN))
```
mpeterss: This does not have to be a server error. It could also be client error 404 Not Found

prasanthu: Not found is handled in normal path( as NotFoundError), exception is treated as bad gateway

PR #412, opened 2018-12-04
Feature/jumio integration

prasanthu:


admin-api/src/main/kotlin/org/ostelco/prime/admin/api/JumioKYC.kt
```diff
@@ -0,0 +1,42 @@
...
+import javax.ws.rs.core.Context
+import javax.ws.rs.core.HttpHeaders
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response
+
+/**
+ * Resource used to handle the eKYC related REST calls.
+ */
+@Path("/ekyc/callback")
+class KYCResource {
```
kmmatwork: A bit short name that does not explains what it is maybe? ` JumioKYCResource` could be an alternative, as it mentions the 3rd party company name. No big deal though.

PR #424, opened 2018-12-18
Feature/sim inventory:  Certificates and gradle

la3lma:
* Generate  crypto-certs and keystores to be used in upcoming integration test.
* Add gradle build scripts, but keep the maven scripts until we're certain that the gradle thingies does all they should do.


sim-administration/certificate-authority-simulated/circle-simulated.sh
```diff
@@ -36,14 +40,26 @@ done
...
+
 ARTEFACT_ROOT=crypto-artefacts
-if [[ -r "$ARTEFACT_ROOT" ]] ; then 
-   rm -f $ARTEFACT_ROOT
+
+
+# Reset if requested
+if [[ -n "$RESET_CRYPTO_ARTEFACTS" ]] ; then 
+    if [[ -f "$ARTEFACT_ROOT" ]] ; then 
+	rm -f $ARTEFACT_ROOT
```
kmmatwork: Will fail in the presence of subdirectories. I guess that '-rf' would better...

sim-administration/certificate-authority-simulated/circle-simulated.sh
```diff
@@ -36,14 +40,26 @@ done
...
+SIM_MANAGER="sim-mgr"
+SMDPPLUS="sm-dp-plus"
+
+ACTORS="$SIM_MANAGER $SMDPPLUS"
 
-ACTORS="sim-mgr sm-dp-plus"
 for  x in $ACTORS ; do 
-  mkdir -p "$ARTEFACT_ROOT/$x"
+    if [[ ! -d "$ARTEFACT_ROOT/$x" ]] ; then 
+        mkdir -p "$ARTEFACT_ROOT/$x"
```
kmmatwork: Minor. The `if [[` test is actually not needed as the '-p' option to mkdir turns the command into a noop if the directory already exists.

PR #426, opened 2019-01-03
Feature/ccr i unknown user

mpeterss:
If the user is not in our DB we should send the correct DIAMETER response code so that the P-GW knows this user is not allowed access.

This means an update to Prime to be able to send error codes both an the base diameter and mscc level.


mpeterss:
This is now deployed to dev

PR #429, opened 2019-01-09
Feature/update client api auth0 domain

havard024:
Updated the auth0 domain to use the redotter-client-dev auth0 tenant in development and the redotter-client auth0 tenant in production.

I've added a custom domain for both tenants:
client-auth-dev.redotter.sg
client-auth.redotter.sg

If anyone has suggestion for better naming or want to follow some kind of naming scheme, I can just create new custom domains for redotter.sg and update the PR.


havard024:
I found another reference to the auth0 domain `ostelco.eu.auth0.com` in [ostelco/prime/auth/OAuthAuthenticator.kt](https://github.com/ostelco/ostelco-core/blob/dc1019ba7a57330fa504d3390f9faa0a46e80497/prime-modules/src/main/kotlin/org/ostelco/prime/auth/OAuthAuthenticator.kt) is this something that is used with the client api? If so, that has to be updated as well.

Should we also update the client api in `new-dev` and `new-prod` what are they used for?

PR #447, opened 2019-02-07
removing hard-coded credential file names

sami-alajrami:
and project ID.


sami-alajrami:
this is covered by #454

PR #448, opened 2019-02-07
Make grpc streams threadsafe

mpeterss:
jDiameter will use many threads and since the
grpc streams are not thread safe this will
use producer/consumer to synch writes to the
grpc stream.


mpeterss:
I need to use synchronized on the queue for wait and notify. Which makes this not optimal. Suggestions are welcome for better ways to solve this :)

PR #466, opened 2019-02-19
Feature/sim inventory

la3lma:
Updating scripts for full regression tests.


run-full-regression-test.sh
```diff
@@ -0,0 +1,84 @@
...
+##
+##  Utility script for running regression tests on a developer workstation.
+##  It is an automation of the procedure described in the docs/TEST.md file.
+##  It is a very good idea to run this script frequently, and at least before
+##  committing something in a PR.
+##
+
+
+DEPS="gradle openssl"
+for dep in DEPS ; do
```
la3lma: This is a bug, it's fixed in a branch that's already being prepared.  It's minor so I'll let that pass for now.

sim-administration/simmanager/deploy-to-dev.sh
```diff
@@ -1,17 +1,84 @@
...
 
-kubectl config use-context $(kubectl config get-contexts --output name | grep dev-cluster)
+##
+##  Utility script to help build a new  docker image of the sim manager
+##  and to deploy it to  a kubernetes cluster.
+##
 
+<<<<<<< HEAD
+
+DEPENDENCIES="kubectl docker gcloud"
```
la3lma: This is a bad one.  Need to fix this

PR #490, opened 2019-03-06
Fix/prime new deployment

sami-alajrami:
only updates deployment config.


la3lma:
Seems like the CircleCI build isn't terminating, or at least not signalling its termination in a manner that is recognised by GitHub.

sami-alajrami:
my bad @la3lma .. the branch name does not start with feature/... so circleci is not building it at all. I will create a new branch and PR.

PR #494, opened 2019-03-08
Rewrite of SIM manager to use the 'Either' pattern (WIP)

kmmatwork:


kmmatwork:
Cancelling this PR as the branch name is incorrect, causing it not to be pushed to Circle-Ci.

PR #501, opened 2019-03-12
Feature/hss adapter interface

la3lma:
Major refactoring, change name from  hlr to hss, merge with Either-ized version of the sim manager, prepare split into strongly separated HSS adapters running as a separate service in k8.


sim-administration/simmanager/src/integration-test/kotlin/org/ostelco/simcards/admin/SimAdministrationTest.kt
```diff
@@ -199,95 +180,47 @@ class SimAdministrationTest {
     }
 
 
-    @Test
```
kmmatwork: Is it a good idea to remove the SIM tests?

la3lma:
testAllocateNextEsim's functionality is subsumed by the batch allocation mechanism, and there is an integration test for that.

sim-administration/simmanager/src/integration-test/kotlin/org/ostelco/simcards/admin/SimAdministrationTest.kt
la3lma: Removed the test because the resource methods were removed. No longer needed.

PR #502, opened 2019-03-13
PubSub OCS API

vihangpatil:
- Removed Pantel from `run-full-regression-test.sh`.
- Abstracted out Protobuf code in `OCSGW` from `GrpcDataSource` to `ProtobufDataSource`.
- Added `PubSubDataSource` which internally uses `ProtobufDataSource`.
- Removed ESPs from `docker-compose.*` files.
- Retained ESPs in new `docker-compose.esp.yaml` if we ever want to test ESPs.
- Added feature in OCSGW to disable TLS, which is needed for AT without ESPs.


ocsgw/src/main/java/org/ostelco/ocsgw/OcsServer.java
```diff
@@ -101,18 +103,35 @@ void init(Stack stack, AppConfig appConfig) throws IOException {
...
+                        appConfig.getPubSubSubscriptionIdForCcr(),
+                        appConfig.getPubSubSubscriptionIdForActivate());
                 break;
             case DataSourceType.LOCAL:
                 LOG.info("Using LocalDataSource");
                 source = new LocalDataSource();
                 break;
             case DataSourceType.PROXY:
                 LOG.info("Using ProxyDataSource");
-                GrpcDataSource secondary = new GrpcDataSource(appConfig.getGrpcServer(), appConfig.getMetricsServer());
```
mpeterss: We should have a config to set secondary to either grpc or pubsub

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/ProtobufDataSource.kt
```diff
@@ -0,0 +1,169 @@
...
+
+    private val sessionIdMap = ConcurrentHashMap<String, SessionContext>()
+
+    fun handleRequest(context: CreditControlContext): CreditControlRequestInfo? {
+
+        logger.info("[>>] creditControlRequest for {}", context.creditControlRequest.msisdn)
+
+        // FixMe: We should handle conversion errors
+        val creditControlRequestInfo = ProtobufToDiameterConverter.convertRequestToGrpc(context)
+        if (creditControlRequestInfo != null) {
```
mpeterss: This could be simplified to :        

if (creditControlRequestInfo != null) {
            ccrMap[context.sessionId] = context
            addToSessionMap(context)
}
return creditControlRequestInfo

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/PubSubDataSource.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.ostelco.diameter.getLogger
+import org.ostelco.ocs.api.ActivateResponse
+import org.ostelco.ocs.api.CreditControlAnswerInfo
+import org.ostelco.ocsgw.datasource.DataSource
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+
+
+class PubSubDataSource(
+        private val asyncDataSource: ProtobufDataSource,
```
mpeterss: rename val to protobufDataSource ?

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/ProtobufDataSource.kt
```diff
@@ -0,0 +1,169 @@
...
+            ccrMap[context.sessionId] = context
+            addToSessionMap(context)
+            return creditControlRequestInfo
+        }
+        return null
+    }
+
+    fun handleProtobufCcrAnswer(answer: CreditControlAnswerInfo) {
```
mpeterss: the class is already named protobuf so this can be

handleCcrAnswer(

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/ProtobufDataSource.kt
```diff
@@ -0,0 +1,169 @@
...
+                }
+            } else {
+                logger.warn("Missing CreditControlContext for req id {}", answer.requestId)
+            }
+        } catch (e: Exception) {
+            logger.error("Credit-Control-Request failed ", e)
+        }
+    }
+
+    fun handleProtobufActivateResponse(activateResponse : ActivateResponse) {
```
mpeterss: fun handleActivateResponse(activateResponse : ActivateResponse) {

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/PubSubDataSource.kt
```diff
@@ -0,0 +1,130 @@
...
+            // add an asynchronous callback to handle success / failure
+            ApiFutures.addCallback(future, object : ApiFutureCallback<String> {
+
+                override fun onFailure(throwable: Throwable) {
+                    if (throwable is ApiException) {
+                        // details on the API exception
+                        logger.warn("Status code: {}", throwable.statusCode.code)
+                        logger.warn("Retrying: {}", throwable.isRetryable)
+                    }
+                    logger.warn("Error publishing active users list")
```
mpeterss: Fix error message

mpeterss: how should we handle failures. should it be retried ?

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/PubSubDataSource.kt
```diff
@@ -0,0 +1,130 @@
...
+                }
+
+                override fun onSuccess(messageId: String) {
+                    // Once published, returns server-assigned message ids (unique within the topic)
+                    logger.debug(messageId)
```
mpeterss: anyway to know which message succeeded if this is a server assigned message id ?

vihangpatil: This might not be possible.

mpeterss: ok, maybe just remove log line, as it has no context.

vihangpatil: There is no action to be taken for successful submission.  If there was, that would be handled there.  Better to remove it.

mpeterss: 👍

PR #503, opened 2019-03-13
Using Google Distroless Java11 for Prime

vihangpatil:


prime/Dockerfile
```diff
@@ -15,4 +15,7 @@ EXPOSE 8081
 EXPOSE 8082
 EXPOSE 8083
 
-CMD ["/start.sh"]
\ No newline at end of file
+# Uncomment line below if you want to use openjdk
+# ENTRYPOINT ["java", "-Dfile.encoding=UTF-8", "-jar"]
```
vihangpatil: I was not aware that developers somethings login into prime docker container for debugging.
Also, this `Dockerfile` is only for deployment onto K8S.
The `Dockerfile.test` used for AT is still on `openjdk`.
@prasanthu Should it be okay to merge this with the comments on how to revert to openjdk, if needed?

Anyways this is not on my top priority.

prasanthu: I would rather wait for things to stabilize before we switch to this docker image.

prasanthu: `tcpdump`, `google cloud utils` etc are some of the tools I have installed to test. I have even installed some compiler to test the `tink` library ;-)

PR #510, opened 2019-03-20
Added support for quota threshold in CCR

mpeterss:


mpeterss:
Quota-Threshold is now set to 80% of the requested quota.
Quota-Holding-Time is set to 2 hours.

This should ideally be configurable. The configuration values should be shared between ocsgw and Prime though.

PR #514, opened 2019-03-22
Generic Datastore. (WIP)

vihangpatil:


vihangpatil:
I was thinking of a generic datastore library which will take any class as an input.
It will then use the class name to be used as Datastore "kind".
The field names will then be fields of that kind. The names can be overridden using `@JsonProperty` annotation. 

Check the test case, where I have created a sample Kotlin data class, which is stored and then retrieved from datastore.
It uses the same object to map conversion that is used in firebase-store and neo4j-store.

This library can then be extended to use more features from Datastore, such as a hierarchy of Kind etc.

data-store/src/main/kotlin/org/ostelco/prime/store/datastore/Schema.kt
```diff
@@ -0,0 +1,100 @@
...
+        }.service
+        keyFactory = datastore.newKeyFactory().setKind(entityClass.name)
+    }
+
+    fun fetch(key: Key): T {
+        val fullEntity = datastore.fetch(key).single()
+        val map = fullEntity
+                .names
+                .map { name ->
+                    val value = when (entityClass.getDeclaredField(name).type) {
```
kmmatwork: We will also need some annotation or similar that can be used to mark that a specific value should be "excluded from being indexed".

F.ex. for the String type if indexed then the max size of the string is 1500. If not indexed then the size can be longer. To mark that a string should not be indexed, the value must be constructed using a builder. Example:

```
StringValue.newBuilder(text)
          .setExcludeFromIndexes(true)
          .build()
```
Similar for other datastore supported values.

kmmatwork: Suggest that we merge this changeset as it is, and then we'll add additional functionality as needed?

data-store/src/main/kotlin/org/ostelco/prime/store/datastore/Schema.kt
```diff
@@ -0,0 +1,100 @@
...
+                        Timestamp::class.java -> fullEntity.getTimestamp(name)
+                        else -> null
+                    }
+                    Pair(name, value)
+                }
+                .toMap()
+        return objectMapper.convertValue(map, entityClass)
+    }
+
+    fun add(t: T): Key {
```
kmmatwork: Must handle the value set to `null` case too I think? One way of supporting this could be: 
```
    fun add(t: T): Key {
        // convert object to map of (field name, field value)
        val map: Map<String, Any?> = objectMapper.convertValue(t, object : TypeReference<Map<String, Any?>>() {})

        // Entity Builder
        val entity = FullEntity.newBuilder(keyFactory.newKey())

        // for each field, call appropriate setter
        map.forEach { key, value ->
            if (value == null) {
                entity.set(key, NullValue.of())
            } else {
                when (value) {
                    is Key -> entity.set(key, value)
                    is Long -> entity.set(key, value)
                    is Blob -> entity.set(key, value)
                    is Double -> entity.set(key, value)
                    is Boolean -> entity.set(key, value)
                    is LatLng -> entity.set(key, value)
                    is String -> entity.set(key, value)
                    is Timestamp -> entity.set(key, value)
                }
            }
        }
        return datastore.add(entity.build()).key
    }
```

PR #521, opened 2019-03-28
Feature/multiple rating groups

mpeterss:


mpeterss:
Still not sure I did not break the coroutine

docker-compose.yaml
```diff
@@ -56,6 +56,7 @@ services:
       - PUBSUB_CCA_TOPIC_ID=ocs-cca
       - PUBSUB_CCA_SUBSCRIPTION_ID=ocsgw-cca-sub
       - PUBSUB_ACTIVATE_SUBSCRIPTION_ID=ocsgw-activate-sub
+      - OCS_DATASOURCE_TYPE=Proxy
```
vihangpatil: Does this mean we will not involve OCS in Prime during AT?

mpeterss: It will use the same setup as we use in dev/prod.  CRR - Init and Terminate will go to Prime/OCS. Update will be local loopback but allse sent to Prime/OCS.

ocs-ktc/src/main/kotlin/org/ostelco/prime/ocs/core/OnlineCharging.kt
```diff
@@ -30,91 +34,111 @@ object OnlineCharging : OcsAsyncRequestConsumer {
...
-                                        && mscc.requested.totalOctets > 0) {
-
-                                    if (consumptionResult.granted < mscc.requested.totalOctets) {
-                                        responseMscc.finalUnitIndication = FinalUnitIndication.newBuilder()
-                                                .setFinalUnitAction(FinalUnitAction.TERMINATE)
-                                                .setIsSet(true)
-                                                .build()
+                val doneSignal = CountDownLatch(request.msccList.size)
+
+                request.msccList.forEach { mscc ->
```
vihangpatil: forEach is not async.  So, we might not need CountDownLatch.

docker-compose.yaml
vihangpatil: Oh, sorry.  This is Proxy, not Local!

ocs-ktc/src/main/kotlin/org/ostelco/prime/ocs/core/OnlineCharging.kt
mpeterss: ok, I added it since the storage access is async

PR #523, opened 2019-03-30
Fix: MyInfo

vihangpatil:


prime/config/config.yaml
```diff
@@ -34,6 +34,7 @@ modules:
...
+      myInfoPersonDataAttributes: "name,sex,race,nationality,dob,email,mobileno,regadd,housingtype,hdbtype,marital,edulevel,assessableincome,ownerprivate,assessyear,cpfcontributions,cpfbalances"
```
prasanthu: I was only asking for `name,nationality,dob,email,mobileno,regadd` from client. Do we really need all other attributes ?

vihangpatil: Yes, we do not need all the attributes when we use our MyInfo API Client ID.  Currently we are using MyInfo API client ID of the demo web app, and hence keeping all the values like Person Attributes and Callback URL same as those in Demo web app.

PR #529, opened 2019-04-02
Remove keepalive timer on grpc channelbuilder

mpeterss:
It seems that the best option is to keep these settings
on the default and have our own ping sent on the stream
instead.


mpeterss:
This does not fix the issue. But setting these values does not seem to help at all. We instead see some issues with "to many pings" when these are set.

PR #537, opened 2019-04-04
Good Bye Pantel

vihangpatil:
Removing legacy kubernetes configuration and deployment scripts.
We now deploy using CircleCI and helm.


vihangpatil:
@prasanthu please check if `exporter` is also moved to helm.  If so, remove `exporter/deploy`.

vihangpatil:
@la3lma please check if `bq-metrics-extractor` is also moved to K8s.  If so, then remove `bq-metrics-extractor/cronjob`.

vihangpatil:
@prasanthu please check if `scaninfo-shredder` is moved to helm.  If so, then remove `scaninfo-shredder/cronjob`.

prasanthu:
> @prasanthu please check if `exporter` is also moved to helm. If so, then in `exporter/deploy`, remove k8s config file and update deployment scripts to use `helm` instead of `kubectl`.

@vihangpatil @rmz Whats is the point of `exporter` now, the whole strategy around analytics is changed/changing.

vihangpatil:
We do not need the `psudeo id` joining part of `exporter`.  But, I'm not sure if there is some logic which will be still useful.  If not, then it's better to delete the module.  It will still be there in git history if we need to refer it again.

vihangpatil:
I'm merging this PR.  If more such files are to be removed, let's do in separate PR.  Better to merge this soon before anyone does more changes to these files.

PR #564, opened 2019-04-16
Fix Neo4j DB transaction/connection pool leak

vihangpatil:


prasanthu:
I have added some Logs to the NRIC verification API,  Should I remove them ?

PR #565, opened 2019-04-16
SIM Profile status updates.

vihangpatil:
SIM Profile status updates using a callback lambda.
Sync SIM Profiles status with SIM Manager using Admin API.


vihangpatil:
1. Are we syncing Neo4j with Sim-manager ?
A: Yes. It also has the additional code to update status in Neo4j when status changes in SIM Manager.

Who will use this API? Should this be added to the UI?
A: No, not needed.  Since the status update code was missing, the SIM Manager and Neo4j status are out of sync.  So, this code is added to sync them.  I might remove it or change it to cronjob to "check and alert if there is a mismatch but do not sync".
 
Shouldn't we add an entry to openAPI spec (houston??)
A:  Not needed since we are not adding it to Admin UI.

PR #567, opened 2019-04-23
Updating production deployment config

sami-alajrami:


.circleci/config.yml
```diff
@@ -379,9 +379,9 @@ workflows:
       - deploy-to-dev:
           requires:
             - update-dev-endpoints
-  #    - create-PR-to-master:
-  #        requires:
-  #          - deploy-to-dev
+      - create-PR-to-master:
+          requires:
+            - deploy-to-dev
```
vihangpatil: I remember this was failing because CircleCI does not have permissions to create a new PR.

PR #568, opened 2019-04-24
Fixes incorrect error type on incorrect parameters in Stripe requests

kmmatwork:


kmmatwork:
Small change so OK without review...

PR #569, opened 2019-04-24
Updates creation of 'customer' in Stripe with Prime 'id' and 'email'

kmmatwork:
Previously a 'customer' in Stripe was created with 'email' set
to 'uuid@ostelco.org' and 'id' set to an id created by Stripe.
This update changes this such that the 'id' in Stripe now is set
to 'uuid' and 'email' to 'contactEmail'.


mpeterss:
LGTM. I do not know the reason by the previous implementation though. But with this we can fetch users on the correct email and uuid. Problem might be that we need to synch email if they change/update email so that it change in stripe as well.

kmmatwork:
> Problem might be that we need to synch email if they change/update email so that it change in stripe as well.

Yes, I expect that there will be an update later that updates the information in Stripe whenever the user updates its contact information + that more information, such as name etc. also gets stored to Stripe - as this information might be useful when generating receipts/invoices from Stripe.

PR #577, opened 2019-04-26
Fix currency values in predefined products

kmmatwork:


kmmatwork:
Will merge this when tests has passed...

PR #580, opened 2019-04-28
Fixes incorrect way of writingi the 'kr' value in predefined products

kmmatwork:


kmmatwork:
Fyi., will merge when tests pass.

PR #584, opened 2019-04-30
Harden esim management based on last couple of weeks bug-search.

la3lma:


sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
```diff
@@ -72,19 +68,27 @@ class PreallocateProfilesTask(
...
+            simInventoryDAO.findNextNonProvisionedSimProfileForHss(hssId = hssEntry.id, profile = simProfileName)
                     .flatMap { simEntry ->
-                        preProvisionSimProfile(hlrEntry, simEntry)
+                        preProvisionSimProfile(hssEntry, simEntry)
                                 .mapLeft {
                                     logger.error("Preallocation of SIM ICCID {} failed with error: {}}",
                                             simEntry.iccid, it.description)
+
+                                    // XXX This looks weird, get review on it (and the code above)
+                                    simInventoryDAO.setSmDpPlusState(simEntry.id!!, SmDpPlusState.ALLOCATION_FAILED)
```
vihangpatil: If `id` is never null, then no need to have it optional.  If it is nullable, then handle it by either wrapping it inside if-not-null check.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
```diff
@@ -14,11 +14,7 @@ import org.ostelco.prime.simmanager.NotFoundError
 import org.ostelco.prime.simmanager.SimManagerError
 import org.ostelco.simcards.hss.HssEntry
 import org.ostelco.simcards.hss.SimManagerToHssDispatcherAdapter
-import org.ostelco.simcards.inventory.HssState
-import org.ostelco.simcards.inventory.SimEntry
-import org.ostelco.simcards.inventory.SimInventoryDAO
-import org.ostelco.simcards.inventory.SimProfileKeyStatistics
-import org.slf4j.LoggerFactory
+import org.ostelco.simcards.inventory.*
```
kmmatwork: Explicit imports are better (I think)

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
kmmatwork: The '!!' will cause a KotlinNullException if the variable is not set. Maybe some pre check before using the variable? Or better, ensuring that `id` is always set if `simEntry` exists.

kmmatwork: Would perfer that "provision-state" is used for marking invalid entires. That is extend the `ProvisionState` enum in `SimInventoryDAO` with the "ALLOCATION_FAILED" field and use the `dao.setProvisionState()` call. In this way we will have only "official" SmDpPlus states. Plus, keeping the SmDpPlus state unchanged will tell us at which state the entry was in when the allocation for some reason failed.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
```diff
@@ -101,18 +105,18 @@ class PreallocateProfilesTask(
...
                         .bind()
 
-                for (entry in hssEntries) {
-                    val profiles: Collection<String> = simInventoryDAO.getProfileNamesForHssById(entry.id)
+                for (hssEntry in hssEntries) {
+                    val simProfileNames: Collection<String> = simInventoryDAO.getProfileNamesForHssById(hssEntry.id)
                             .bind()
-                    for (profile in profiles) {
-                        val profileStats = simInventoryDAO.getProfileStats(entry.id, profile)
+                    for (simProfileName in simProfileNames) {
```
kmmatwork: Cosmetic - have started to prefer `forEach ...`

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SimManager.kt
```diff
@@ -42,17 +36,25 @@ object SimManagerSingleton : SimManager {
...
+        return when (smdpPlusState) {
+            SmDpPlusState.AVAILABLE -> SimProfileStatus.NOT_READY
+            SmDpPlusState.ALLOCATED -> SimProfileStatus.NOT_READY
+            SmDpPlusState.CONFIRMED -> SimProfileStatus.NOT_READY
+            SmDpPlusState.RELEASED -> SimProfileStatus.AVAILABLE_FOR_DOWNLOAD
+            SmDpPlusState.DOWNLOADED -> SimProfileStatus.DOWNLOADED
+            SmDpPlusState.INSTALLED -> SimProfileStatus.INSTALLED
+            SmDpPlusState.ENABLED -> SimProfileStatus.ENABLED
+            SmDpPlusState.ALLOCATION_FAILED -> SimProfileStatus.NOT_READY
+        }
```
kmmatwork: No `else` part? What happens if there are no match, f.ex. in case of incomplete code update?

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/inventory/SimInventoryDAO.kt
```diff
@@ -40,6 +40,9 @@ enum class SmDpPlusState {
     DOWNLOADED,
     INSTALLED,
     ENABLED,
+
+    /* This is not a legal ES2+ state, and it means that allocation in the SM-DP+ failed */
+    ALLOCATION_FAILED
```
kmmatwork: Move to `ProvisionState` (see comment further up).

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/inventory/SimInventoryDAO.kt
```diff
@@ -211,7 +214,7 @@ class SimInventoryDAO(private val db: SimInventoryDBWrapperImpl) : SimInventoryD
     //
 
     override fun insertAll(entries: Iterator<SimEntry>): Either<SimManagerError, Unit> =
-        db.insertAll(entries)
+            db.insertAll(entries)
```
kmmatwork: A TAB? Prefer space myself - except in Makefiles

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
la3lma: The !! is necessary since the class permits null ID values, so that instances can be created before the datbase ID is known. When reading from the database this will never happen (obviously), but the type system isn't informed about this.    

In this case an explicit test for null values of the simEntry.id field  could make the "!!" redundant, but I actually don't know how to do that using Arrow.  Any suggestions?

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SimManager.kt
la3lma: It fails during compilation .-)

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/inventory/SimInventoryDAO.kt
la3lma: *blush*

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
la3lma: ok

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
la3lma: Fixed, also updated my intellij preferences to don't do it again.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
la3lma: I added some comments essentially describing the above then I'll let it pass.   Getting rid of "!!" is a bigger issue anyway. However, this is also an issue where my understanding of control flow in arrow gets in the way of fixing an issue.

la3lma: I found a way (a kludgy if statement :-)

la3lma:
Fixed all things commented on by reviewers.    Merging.

PR #598, opened 2019-05-07
RAR message needs to be sent with vendor Id set for 3GPP

mpeterss:


mpeterss:
This turned out to be not needed for RAR. Will keep branch around it there is other use for it.

PR #616, opened 2019-05-13
Feature/import sim batch for hlr pre activated msisdns

la3lma:
Unit and integration tests added for import script that is now modified to set the hss state from a query parameter.


build-all.sh
```diff
@@ -93,7 +93,7 @@ DIRS_THAT_NEEDS_SERVICE_ACCOUNT_CONFIGS=" \
   bq-metrics-extractor/config \
   auth-server/config prime/config"
 
-SERVICE_ACCOUNT_MD5="b8d22f87b55431ebfa4b68c16d7dc037"
+SERVICE_ACCOUNT_MD5="c54b903790340dd9365fa59fce3ad8e2"
```
la3lma: This is an unrelated change.  It's for making the "build-all.sh" script recognize the latest prime-service-account.json instancde.

sim-administration/simmanager/src/integration-test/kotlin/org/ostelco/simcards/admin/SimAdministrationTest.kt
```diff
@@ -35,6 +44,11 @@ class SimAdministrationTest {
         private lateinit var jdbi: Jdbi
         private lateinit var client: Client
 
+        // ICCID of rirst SIM in sample-sim-batch.csv
```
prasanthu: spelling mistake `rirst`

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/ArrowUtils.kt
```diff
@@ -5,6 +5,5 @@ import arrow.core.Either
 
 /**
  * TODO: 1. Document this method
- *       2. Move it into a convenience libary for Arrow utilities.
  */
 fun <L,R, R2> Either<L, R>.mapRight(f:(R)->R2): Either<L, R2> = this.map(f)
```
prasanthu: No newline

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/inventory/SimInventoryResource.kt
```diff
@@ -111,15 +112,21 @@ class SimInventoryResource(private val api: SimInventoryApi) {
...
-                            {
-                                error("Failed to upload batch with SIM profiles for HSS ${hss} and SIM profile vendor ${simVendor}",
-                                        ApiErrorCode.FAILED_TO_IMPORT_BATCH, it)
-                            },
-                            { Response.status(Response.Status.OK).entity(asJson(it)) }
-                    ).build()
+            @QueryParam("initialHssState") initialHssStateArg: HssState?,
+            csvInputStream: InputStream): Response {
+
+        val initialHssState = if (initialHssStateArg==null) HssState.NOT_ACTIVATED else initialHssStateArg
```
kmmatwork: Since this is at input I guess that it could be an idea to check whether the `initialHssStateArg` value, if set, is a valid value? That is matches one of the entries in the `HssState` enum.

sim-administration/simmanager/src/integration-test/kotlin/org/ostelco/simcards/admin/SimAdministrationTest.kt
la3lma: fixed

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/inventory/SimInventoryResource.kt
la3lma: It's implicitly checked by the input filter for the applet.   If it's set it will automatically be translated to a  legal value otherwise a HTTP 500 will be thrown (will add explicit test for that).  I don't know if it's possible to test for illegal query parameters to be set or if they fail silently.  I will check.  It would be better to fail than to fail silently, because silent failure of this kind will actually import the data but with the wrong value for initial hss state set.  Good catch.

la3lma: fixed now.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/ArrowUtils.kt
la3lma: Fixed

PR #617, opened 2019-05-14
Remove MSCC in terminate

mpeterss:
As there was no Requested-Service-Units in the CCR-T then
adding MSCC in CCA-T seems a bit strange. RFC is not specific
in this point so it needs to be tested.


vihangpatil:
CCR-Terminate does send `used` ServiceInfo, even if the value for `requested` serviceInfo will be zero.  That serviceInfo still needs MSCC, right?

mpeterss:
yes, the CCR-T will send the Used-Service-Units in the MSCC in the request.
What this change is that in the CCA-T we used to send back a MSCC with zero grant. Now this will remove it.

As there was no Requested-Service-Units in the CCR-t is seems a bit strange to send a grant in the answer. This was here because this is how the old implementation used to do it.

PR #618, opened 2019-05-14
Feature/reserve 0000 and 9999 ending numbers

la3lma:
MSISDNs ending in '0000' or '9999' are "golden", and should  by default not be allocated.  Some jurisdictions requires special handling of these numbers, and now we open up for that, by leaving these numbers marked as "reserved" in our sim profile database.


la3lma:
This PR is mostly the https://github.com/ostelco/ostelco-core/pull/616 PR, with all bugs fixed and all comments taken into account, plus golden number management.

PR #660, opened 2019-06-03
Removes dependency on 'region' for determining 'tax-region' to apply and which offer to provide when buying a plan

kmmatwork:


neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -1094,30 +1094,22 @@ object Neo4jStoreSingleton : GraphStore {
...
-                        }
-                        .mapLeft {
-                            BadGatewayError("No region found for ${customer.id}", error = it)
-                        }
-                        .bind()
+                /* Fetch 'tax' id to be applied. */
+                val taxRegionId = if (product.payment.containsKey("taxRegionId"))
+                    product.payment["taxRegionId"]
+                else if (product.price.amount > 0)
+                    BadGatewayError("Product ${product.sku} has price ${product.price.amount} ${product.price.currency} but no 'taxRegionId' set")
```
vihangpatil: In next PR handle the case where the Product has price.amount > 0 but does not have a Tax associated with it.

PR #661, opened 2019-06-03
Use cloud logger in ocsgw

mpeterss:


vihangpatil:
Originally we had this Google Cloud Logger when OCSGW was NOT running in GCP.
But now since it is already running in GCP, it just has to log to stdout in json format.
@mpeterss why are we adding the Cloud logger back?

ocsgw/build.gradle
```diff
@@ -25,6 +25,8 @@ dependencies {
     implementation 'ch.qos.logback.contrib:logback-json-classic:0.1.5'
     implementation 'ch.qos.logback.contrib:logback-jackson:0.1.5'
 
+    compile 'com.google.cloud:google-cloud-logging-logback:0.92.0-alpha'
```
vihangpatil: Latest version `0.94.0-alpha` can be used.

ocsgw/config/logback.dev.xml
```diff
@@ -1,26 +1,31 @@
...
     <encoder>
       <pattern>%d{dd MMM yyyy HH:mm:ss,SSS} %-5p %c{1} - %m%n</pattern>
     </encoder>
   </appender>
 
+  <appender name="CLOUD" class="com.google.cloud.logging.logback.LoggingAppender">
+    <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+      <level>DEBUG</level>
+    </filter>
+    <log>application.log</log>
```
vihangpatil: `application.log` is the default value.  Better to change it to `ocsgw`.
Then we can use this expression to filter in Stackdriver logging:
`logName="projects/<gcp project>/logs/ocsgw"`

mpeterss:
@vihangpatil My problem was that when using STDOUT logger it is not setting Level correctly in the message. Also when I was logging the AVPs i put a log message with linebreaks in it. When using stdout it split each newline to a new log message. 

Google recommendation for GCE seems to be to use CLOUD logger.
https://cloud.google.com/logging/docs/setup/java
With <resourceType>gce_instance</resourceType> (This is autodetected)

But I am not sure what is correct here.

mpeterss:
The cloud logger does not support encoders.

-ERROR in ch.qos.logback.core.joran.spi.Interpreter@11:14 - no applicable action for [encoder], current ElementPath  is [[configuration][appender][encoder]]

and by default it does not add classname to log message. This is what a log message looks like : 

'2019-06-05 13:28:17.166 CEST Created Credit-Control-Answer'

This is not optimal

mpeterss:
I will merge this so we have working logging until we get help from Google how to solve this proper.

PR #665, opened 2019-06-06
Bugfix/capitalization in bssname issue

la3lma:


la3lma:
Should be called feature/ not bugfix/.  Deleting branch and creating new properly named one.

PR #670, opened 2019-06-10
[WIP] Json formatted logging for ocsgw

vihangpatil:


vihangpatil:
@mpeterss I have changed both files - `logback.xml` and `logback.dev.xml` to do json formatted logging.  But we need normal logging for acceptance tests.  I'm not sure which file is used for AT and which one is used in `dev`/`prod` env.

vihangpatil:
@mpeterss The solution can be verified by `docker-compose up --build` in `ocsgw`.

mpeterss (changes requested):
We will hold off with this until we get feedback from Google how this is supposed to work. For now I have set the nodes to use the cloud logger.

mpeterss:
This is broken in GCP
Please star the ticket
https://issuetracker.google.com/137517429

vihangpatil:
This will not work due to the limitation in GCP. Closing the PR.

PR #671, opened 2019-06-11
Feature/dummy hss

la3lma:


sim-administration/hss-adapter/src/integration-test/resources/hss-adapter-config.yaml
```diff
@@ -4,11 +4,13 @@ logging:
 
 hlrs:
   - name: Foo
+    type: WG2
```
kmmatwork: Maybe "vendor" would be a better name? But this of course depends on context...

la3lma: fixed.

PR #696, opened 2019-06-20
Update direct deploy and canary setup

sami-alajrami:
This PR uses a new prime helm chart version (1.0.0) with updated deployment config. When merged, it will:

-  Enable deploying prime-direct with the same endpoints as dev and with header-based traffic filtering. (it is also possible to use weights).
- allow configuring canary settings for individual services.
- Split stable and canary deploys in prod so that they are independently deployable. 
- Introduce another manual step in the pipeline to promote canary to stable in prod.
- Schedule prime pods on specific nodes in dev and prod clusters.
- Reduce differences between prime-direct/dev/canary/stable prod deployment config and set the same resource requests for all.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -269,7 +269,7 @@ jobs:
             /authenticate.bash
             helm repo add ostelco https://storage.googleapis.com/pi-ostelco-helm-charts-repo/
             helm repo update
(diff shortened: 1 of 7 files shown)


vihangpatil (approved):
Thanks a lot @sami-alajrami

PR #709, opened 2019-06-24
Modify prime direct canary

sami-alajrami:
this PR makes it possible to run commands for reconfiguring traffic routing between prime and prime-direct:

- To set all traffic to prime-direct (without headers)
```
./prime/script/change-prime-direct-canary.sh <service name, e.g. api> full-weight
```

- To set all traffic to prime (without headers)
```
./prime/script/change-prime-direct-canary.sh <service name, e.g. api> zero-weight
```

- To set traffic with the header `x-mode=prime-direct` to prime direct:
```
./prime/script/change-prime-direct-canary.sh <service name, e.g. api> header
```

### DIFF
--- a/.circleci/prime-dev-values.yaml
+++ b/.circleci/prime-dev-values.yaml
@@ -136,7 +136,7 @@ prime:
     - 8083 
   resources: 
     requests:
-      cpu: 300m
+      cpu: 100m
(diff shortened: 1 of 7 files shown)


sami-alajrami:
LGTM

PR #731, opened 2019-07-02
Feature/fixing double bangs

la3lma:


la3lma:
@vihangpatil  and @kmmatwork : I would like review on this one.  The4re are a bunch of things that I've done that are probably not idiomatic Kotlin, and as our resident language lawyer, I'd like to hear your reasoned opinion on alternative ways of fixing the issues I have fixed so far.

PR #733, opened 2019-07-03
Added sim-manager to AT

vihangpatil:


kmmatwork (approved):
LGTM.

Also supports renaming of `sm-dp-plus` to `sm-dp-plus-emulator` or something similar.

Issue #716, opened 2019-06-25
[Prime] Low balance notifications is no longer sent


mpeterss:
After the refactor to ocs-ktc there is no longer any notifications sent to handsets on low balance.

mpeterss:
https://github.com/ostelco/ostelco-core/pull/758

mpeterss:
https://github.com/ostelco/ostelco-core/pull/761

mpeterss:
Fixed

PR #742, opened 2019-07-11
Moved rating definitions to config

mpeterss:


mpeterss:
Issue https://github.com/ostelco/ostelco-core/issues/716

Issue #754, opened 2019-07-17
[Prime] Activation no longer sent to P-GW on topup


mpeterss:
Since the refactor to ocs-kt the activation message is no longer sent to OCS gateway on topup. This means that the Re-Auth Request is never sent to the P-GW, hence the user is no longer auto enabled again if they topup a zero balance bundle.

This should be fixed in Prime and only send activation messages in the case where the balance goes from zero.

mpeterss:
#755 
#774

PR #755, opened 2019-07-17
Send activation messages when a user topup bundle

mpeterss:


mpeterss:
Fixing issue : https://github.com/ostelco/ostelco-core/issues/754

PR #758, opened 2019-07-18
Add back notifications on low balance

mpeterss:


mpeterss:
Fixing issue https://github.com/ostelco/ostelco-core/issues/716

PR #761, opened 2019-07-19
Use costumer id not msisdn to lookup notification tooken

mpeterss:


mpeterss:
https://github.com/ostelco/ostelco-core/pull/761

PR #771, opened 2019-07-25
Feature/expose legacy api to graphql schema

havard024:
Updated GraphQL schema ~~to reflect the OpenAPI specification for the legacy API.~~ to expose more of the legacy endpoints.

Missing mutations before all legacy APIs have been converted
- loadSingpassInfo
- loadMyInfoConfig
- updateEKYCProfile
- stripeEphemeralKey

Considerations
- Can we avoid the double folds, or is that an ok way to solve it?
- We should, not right now, remove boilerplate to make it easier to expose new endpoints
- We should make sure that all IDs are stored in the same field, like the id field, I've tried to model this by having a Node interface in the GraphQL schema, and use the interface for all types that has a an id. The ID should be global unique and is important for how we use the graphql client libraries.

Errors
- We need to expose the error type in such a way that the client can use it as before
- Errors is a list of errors, how do we handle this on client side, can we guarantee that there only will be one error in some cases?
- GraphQL separates network errors from other errors. Network Errors are any response where the http status code is not in the 2xx range. Other errors is returned with data, you can get partial data, no data all data with no errors, 1 error or multiple errors. This is up to us to design how we want this to work.

Tests
We need these, there are none...

Example of a default error (if something wrong happens that we don't catch or handle) like an exception is thrown while handling the request. (This can be handled in CustomDataFetcherExceptionHandler inside DataFetchers.kt)
```
{
    "errors": [
        {
            "message": "Exception while fetching data (/createAddress) : null",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createAddress"
            ],
            "extensions": {
                "classification": "DataFetchingException"
            }
        }
    ],
    "data": {
        "createAddress": null
    }
}
```

Example of an error where we add custom attributes to the error message
```
{
    "errors": [
        {
            "message": "Identity - havard+test@redotter.sg already exists.",
            "locations": [],
            "extensions": {
                "id": "havard+test@redotter.sg",
                "type": "Identity",
                "classification": "DataFetchingException"
            }
        }
    ],
    "data": {
        "createCustomer": null
    }
}
```


graphql/src/main/kotlin/org/ostelco/prime/graphql/DataFetchers.kt
```diff
@@ -4,12 +4,58 @@ import com.fasterxml.jackson.core.type.TypeReference
...
+            .fold({
+                throw Exception(it.message)
+            }, {
+                clientDataSource.getCustomer(identity)
+                        .map { customer ->
+                            ret = objectMapper.convertValue<Map<String, Any>>(customer, object : TypeReference<Map<String, Any>>() {})
+                        }
+            })
+
+        return ret!!
```
havard024: Would be great if we could return the return value without having to force unwrap an optional.

prasanthu:
Stale PR, closing it for now.

PR #772, opened 2019-07-25
Remove null values from payment source listing

kmmatwork:


kmmatwork:
Applied as part of an another PR.

PR #801, opened 2019-08-28
Lowercases currency before doing checks in payment trancsaction monitor

kmmatwork:


vihangpatil (approved):
LGTM!. Nice to remove redundant `map { it }` as well.

PR #821, opened 2019-09-11
Sim manager refactoring

la3lma:
1. Fix the ES2+ ping protocol used to check if the SM-DP+ in the other end is answering requests.
2. One round of refactoring of ProfileVendorAdapter.kt.  This round of refactoring consists of 
    a) Introducing a set of methods that factor out a lot of repeated code.
    b) Removing some functionality (in particular with respect to eid), that is not necessary for our current usecases, and maybe also not correct.  All in all, this simplification makes the code much more compact and easy to read.

### DIFF
--- a/sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
+++ b/sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
@@ -68,7 +68,7 @@ class SmdpPlusHealthceck(
 
                     val profileVendorAdaptorList = vendorsRaw.bind()
 
-                    for (profileVendor in profileVendorAdaptorList) {
+                    loopOverAllProfileVendors@ for (profileVendor in profileVendorAdaptorList) {
                         logger.info("Processing vendor: $profileVendor")
                         val currentConfig: ProfileVendorConfig? =
                                 profileVendorConfigList.firstOrNull { it.name == profileVendor.name }
@@ -82,9 +82,23 @@ class SmdpPlusHealthceck(
                                 httpClient = httpClient,
                                 config = currentConfig!!
                         )
-                        pingResult.mapLeft { error ->
-                            logger.error("Could not reach SM-DP+ via HTTP PING:", error)
+
+                        // If this was an error, but of an acceptable ("pingOk" == true) kind, meaning that
+                        // the endpoint in the other end actually gave a reasonable answer to a reasonable request,
+                        // indicating that the endpoint is answering requests, then continue to loop over next endpoint,
+                        // otherwise see if there is an error.
+                        when (pingResult) {
+                            is Either.Left -> if (pingResult.a.pingOk) {
+                                continue@loopOverAllProfileVendors
+                            } else {
+                                logger.error("Could not reach SM-DP+ via HTTP PING:", pingResult)
+                            }
+                            is Either.Right -> {
+                            }
                         }
+
+                        // If this was an error, then break according to arrow semantics,
+                        // otherwise just carry on looping over the next endpoint.
                         pingResult.bind()
                     }
                 }.fix()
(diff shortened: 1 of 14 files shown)


sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
```diff
@@ -82,9 +82,23 @@ class SmdpPlusHealthceck(
                                 httpClient = httpClient,
                                 config = currentConfig!!
```
kmmatwork: !! only allowed in tests.... :-)

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
```diff
@@ -68,7 +68,7 @@ class SmdpPlusHealthceck(
 
                     val profileVendorAdaptorList = vendorsRaw.bind()
```
kmmatwork, line 71: Not visible in the diff, but I don't think that wrapping the `checkIfSmdpPlusIsUp` method in a try .. catch is necessary, as all calls returns an Either?

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
la3lma: Fixed in next commit.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SmdpPlusHealthceck.kt
la3lma: Maybe.  Will look into it.  There is a _lot_ of refactoring left in this code.

la3lma: For now adding a TODO, but I will get back to it in a later PR.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapter.kt
```diff
@@ -162,101 +177,112 @@ data class ProfileVendorAdapter(
...
-                        logger.error("SM-DP+ 'order-confirm' message to service {} for ICCID {} failed with status code {} (call-id: {})",
-                                config.name,
-                                simEntry.iccid,
-                                it?.statusLine?.statusCode,
-                                header.functionCallIdentifier)
-                        NotUpdatedError("SM-DP+ 'order-confirm' to ${config.name} failed with code: ${it.statusLine.statusCode}")
-                                .left()
-                    }
+                when {
+                    simEntryId == null -> AdapterError("simEntryId == null").left()
```
kmmatwork: Not sure about this, as the `simEntryId` and `statusId` goes out of scope after the `else` part... 

Maybe there should be a `return` here if either or both of the two variables are null?

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapter.kt
```diff
@@ -162,101 +177,112 @@ data class ProfileVendorAdapter(
...
+            logger.info("SM-DP+ 'order-confirm' message to service {} for ICCID {} completed OK (call-id: {})",
+                    nameOfSimVendor,
+                    iccid,
+                    functionCallIdentifier)
+
+            val simEntryId = simEntry.id
+            val statusMatchingId = status.matchingId
+
+            when {
+                simEntryId == null -> AdapterError("simEntryId == null").left()
```
kmmatwork: A bit "thin" error messages here.

kmmatwork (approved):
LGTM.

But could overall benefit from a bit of rework to make things clearer (see comments). Plus, have not looked at the Go code.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapter.kt
la3lma: Tar det som en kommentar inn i neste runde av refactoring, fikser det ikke akkurat nå (legger det som en "TODO" i koden nå, men pusher ikke til denne PR.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapter.kt
la3lma: Have simplified the code.

PR #822, opened 2019-09-16
Fix bug: Shouldn't report errors produced by proper responses from thee ES2+ stack when in ping mode

la3lma:
tssia


vihangpatil (approved):
Local functions. Interesting!

PR #823, opened 2019-09-16
New endpoint in Houston API to remove a customer

prasanthu:
We are removing the option to remove customer from the client (production app). The admin UI will have an option to remove


vihangpatil (approved):
LGTM!

PR #824, opened 2019-09-16
Dependency updates.

vihangpatil:
Also fixed minor regex in JUnit 5 annotation.

### DIFF
--- a/email-notifier/src/test/kotlin/org/ostelco/prime/notifications/email/MandrillClientTest.kt
+++ b/email-notifier/src/test/kotlin/org/ostelco/prime/notifications/email/MandrillClientTest.kt
@@ -31,7 +31,7 @@ data class TestConfig(val modules: List<PrimeModule>): Configuration()
 
 class MandrillClientTest {
 
-    @EnabledIfEnvironmentVariable(named = "MANDRILL_API_KEY", matches = "*")
+    @EnabledIfEnvironmentVariable(named = "MANDRILL_API_KEY", matches = ".*")
     @Test
     fun `send test email via Mandrill`() {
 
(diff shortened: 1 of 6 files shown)


prasanthu (approved):

email-notifier/src/test/kotlin/org/ostelco/prime/notifications/email/MandrillClientTest.kt
```diff
@@ -31,7 +31,7 @@ data class TestConfig(val modules: List<PrimeModule>): Configuration()
 
 class MandrillClientTest {
 
-    @EnabledIfEnvironmentVariable(named = "MANDRILL_API_KEY", matches = "*")
+    @EnabledIfEnvironmentVariable(named = "MANDRILL_API_KEY", matches = ".*")
```
la3lma, line 34: This is a somewhat unusual annotation, so there should be an accompanying comment stating that the test should only be run I the variable MANDRILL_API_KEY is set to anything non-null

prime/script/start.sh
```diff
@@ -5,5 +5,5 @@ exec java \
     -Dfile.encoding=UTF-8 \
     --add-opens java.base/java.lang=ALL-UNNAMED \
     --add-opens java.base/java.io=ALL-UNNAMED \
-    -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_service=prime,-cprof_service_version=1.60.2,-logtostderr,-minloglevel=2,-cprof_enable_heap_sampling \
+    -agentpath:/opt/cprof/profiler_java_agent.so=-cprof_service=prime,-cprof_service_version=1.61.0,-logtostderr,-minloglevel=2,-cprof_enable_heap_sampling \
```
la3lma, line 8: The versioning is managed with magic strings, and even repeated. In this case there is a version set in prime/build.gradle, and it would be good if it was declared only once.

PR #826, opened 2019-09-17
Changed to Dispathers.IO for neo4j calls

mpeterss:
For DB and network calls it is recommended to use Dispatchers.IO instead of Dispatchers.Default which is optimized to perform CPU-intensive work.


vihangpatil (approved):
Awesome!

PR #827, opened 2019-09-17
Changes in gitignore files

vihangpatil:


mpeterss:
Why do we no longer want to ignore the kotlin script files?

mpeterss (approved):

vihangpatil:
> Why do we no longer want to ignore the kotlin script files?

I am ignoring the whole sub-repo instead.

PR #828, opened 2019-09-17
Feature/sim manager refactoring

la3lma:
Connect ProfileVendorAdapter to the ES2PlusClient, clean up various bits and pieces.  This is still a work in progress, more cleanup necessary before this is in an acceptable state, but using . the protocol that has unit test coverage is an important step, hence this PR.


sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
```diff
@@ -36,6 +35,9 @@ import kotlin.math.min
  * there will be a number of profiles available for quick allocation to customers
  * without having to synchronously wait for a profile to be provisioned by these
  * two.
+ *
+ * TODO:  This is so incredibly complicated, and there is no need for it to be.
+ *        Just fix it!
```
kmmatwork, line 40: As part of this. Noted that the
```
            if (simEntry.id == null) { // TODO: This idiom is _bad_, find something better!
                AdapterError("simEntry.id == null for simEntry = '$simEntry'.").left()
            } else
```
(line 84-86) was already done by the caller, and that these lines therefore can be removed (unless I have overlooked something...). Should be OK as the function is private.

Also on line 150 `return "ok".right()`. Would be better to return `Unit.right()` I think. Plus replace `Any` with `Unit` in line 114.

kmmatwork: Generally I think that readability could be improved if some of the code inside `for` loops was moved out as functions?

sim-administration/es2plus4dropwizard/src/main/kotlin/org/ostelco/sim/es2plus/Es2PlusClient.kt
```diff
@@ -30,90 +30,116 @@ class ES2PlusClient(
         private val useHttps: Boolean = true,
         private val jerseyClient: Client? = null) {
 
+    // TODO: Make a new interface to represent the client we are using, then set a non-nullable field
+    //       (private) to contain one instande that represents either a HttpClient or a jerseyClient
+    //       wrapped in the appropriate protocol logic.   This will get rid of the silly ==null tests
+    //       that riddle this class.
+
+
```
kmmatwork, line 38: Line 29 (outside of changeset, but...). Would it not be better to not allow `httpClient` to be `null`?

kmmatwork: OK. You can use either a Jersey or a http client (line 29 and 31). Maybe just settle for one?

kmmatwork: Agree with the TODO!

sim-administration/es2plus4dropwizard/src/main/kotlin/org/ostelco/sim/es2plus/Es2PlusClient.kt
```diff
@@ -30,90 +30,116 @@ class ES2PlusClient(
...
+        req.entity = StringEntity(payload)
 
-            // Validate returned response
-            val statusCode = result.statusLine.statusCode
-            if (expectedStatusCode != statusCode) {
-                val msg = "Expected return value $expectedStatusCode, but got $statusCode.  Body was \"${result.entity.content}\""
-                throw ES2PlusClientException(msg)
-            }
+        val response: HttpResponse = httpClient.execute(req)
+                ?: throw ES2PlusClientException("Null response from http httpClient")
```
kmmatwork, line 102: A few `throw`statements in the code. Should be rewritten to use `Either` I think. :-)

kmmatwork (approved):
LGTM for my part.

Could still benefit from some refactoring to improve readability etc. But this change is in any case a step in that direction.

sim-administration/es2plus4dropwizard/src/main/kotlin/org/ostelco/sim/es2plus/Es2PlusClient.kt
la3lma: This library does not use either.  A wrapper that does use either has been written in the ProfileVendorAdapter, it may be moved into the client once it is a bit more mature (it's still not compact enough).

sim-administration/es2plus4dropwizard/src/main/kotlin/org/ostelco/sim/es2plus/Es2PlusClient.kt
la3lma: The reason I have both is that the jersey client does not work well with SSL, and the other one does (after a fashion, at least it works). However, the jersey client has excellent support  for unit testing, and most of the original code was written using jersey client,, and it was very productive to do so.   I will look into fixing this one way or another, since having two clients present in this somewhat messy way is not a good practice,  but I will not do that in this PR.

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/PeriodicProvisioningTask.kt
la3lma: I will leave this for now, since it is my intention to refactor this code extensively later, the for loop is only one of many things that will be fixed.  The "Ok" text is another.  The method should return a result that shall be returned to the caller so that the caller can see a summary of  what the task actually accomplished.

la3lma: A rule of thumb is that the level of parenthesis shouldn't be more than three or so, so clearly this method, and several others in this file, break this rule and clear targets for refactoring.

la3lma: I have made TODOs  of these issues in my working copy and will address them in future PRs, possibly even the next one.

la3lma:
Merging, even though this is just the second step of several for refactoring the sim manager code.

PR #829, opened 2019-09-18
Make pubsub messages compliant to analytics ID convention

cyriaquebrousse:
* We now use `customerAnalyticsId` and `subscriptionAnalyticsId`, instead of the various exotic variants from before, such as subscriber ID or MSISDN analytics ID. Doing so, we achieve more clarity.
* We send data consumption messages to Cloud Pubsub in JSON format, rather than Protocol buffer. This way, we are able to use standard Dataflow job template for getting Pubsub messages into Bigquery, and have less code to maintain.

All this is in line with the recent efforts to ramp up analytics. Read more on the data flows [here](https://www.draw.io/?state=%7B%22ids%22:%5B%221T6x6RFtFnVzzXjwmqskPHcf1QLj5N3DU%22%5D,%22action%22:%22open%22,%22userId%22:%22115964574021637508931%22%7D#G1T6x6RFtFnVzzXjwmqskPHcf1QLj5N3DU).


vihangpatil (approved):
LGTM

PR #830, opened 2019-09-18
Porting Gradle scripts from groovy to Kotlin script.

vihangpatil:


build.gradle.kts
```diff
@@ -0,0 +1,105 @@
...
+
+subprojects {
+  tasks.withType<KotlinCompile> {
+    kotlinOptions {
+      jvmTarget = JavaVersion.VERSION_12.majorVersion
+    }
+  }
+}
+
+//tasks.withType<DependencyUpdates> {
```
prasanthu, line 93: Add a TODO, so we remember to enable this later

prasanthu (approved):
Checked few commits. Look OK. If we need to check every line, it will take for ever

PR #833, opened 2019-09-19
Ported Gradle build files from Groovy to Kotlin script.

vihangpatil:


acceptance-tests/build.gradle.kts
```diff
@@ -16,8 +16,8 @@ dependencies {
   val jjwtVersion:String by rootProject.extra
   val jacksonVersion:String by rootProject.extra
 
-  implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion")
-  implementation("org.jetbrains.kotlin:kotlin-test:$kotlinVersion") {
+  implementation(kotlin("stdlib-jdk8"))
```
prasanthu, line 19: Where does it get the version (`$kotlinVersion`). Is it defined in the top level gradle file ?

prasanthu (approved):
Skimmed

acceptance-tests/build.gradle.kts
prasanthu: Found it

PR #836, opened 2019-09-20
Feature/sim manager refactoring

la3lma:
More refactoring.    Remove monad-style code, replace with simpler code.


payment-processor/src/integration-tests/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessorTest.kt
```diff
@@ -74,7 +76,7 @@ class StripePaymentProcessorTest {
     @After
     fun cleanUp() {
         val resultDelete = paymentProcessor.deletePaymentProfile(stripeCustomerId)
-        assertEquals(true, resultDelete.isRight())
+        assertNotFailure(resultDelete)
```
kmmatwork: You can also do something like this:
```
check(resultDelete == ProfileInfo(customerId).right())
```
Will also verify that the data is as expected.

We also have [https://github.com/kotlintest/kotlintest](https://github.com/kotlintest/kotlintest), which could be something to look into?

kmmatwork (approved):
LGTM.

(Would probably have restructured the code quite a bit if it was me writing it though... :-) )

PR #837, opened 2019-09-20
Check for stale pull requests

la3lma:
... essentially a little test for checking how convenient github actions really are


prasanthu:
This is a stale PR, closing it.

PR #838, opened 2019-09-20
Flip a bit to fix a bug

la3lma:


la3lma:
Bugfix.  Please approve then merge (it's an one liner, a flipped bit (only log an error when success is expected, not when it's not)).

vihangpatil (approved):

PR #844, opened 2019-09-24
Improve logging when detecting errors when not being able to ping SM-DP+

la3lma:
In our preprod-environment we have seen intermittent failures, but also that these failures do not report very well what's happening.  This patch adds more details to the error logs, and also cleans up a few minor, and mostly stylistic  issues elsewhere.


sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapterDatum.kt
```diff
@@ -138,11 +138,13 @@ data class ProfileVendorAdapter(
...
-                    } else if (response.profileStatusList == null) {
-                        logAndReturnNotFoundError("Couldn't find any response for query iccidlist='$iccidList'")
                     } else {
-                        val result = response.profileStatusList!! // TODO: Why is this necessary (see if-branch above)
-                        return result.right()
+                        val result = response.profileStatusList
+                        if (result == null) {
+                            logAndReturnNotFoundError("Couldn't find any response for query iccidlist='$iccidList'")
+                        } else {
+                            return result.right()
```
vihangpatil: @la3lma - This is a bug. Remove `return`.  Just `result.right()` should be kept.

vihangpatil (changes requested):

sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/profilevendors/ProfileVendorAdapterDatum.kt
la3lma: fixed

vihangpatil (approved):

kmmatwork (approved):

PR #846, opened 2019-09-25
Refactor of Analytics module

cyriaquebrousse:
Refactoring event generation out of Prime, into Analytics module

* Move event generation logic from prime/Entity to an own Event class in analytics
* Add schemas for BigQuery tables


vihangpatil (approved):
LGTM from high level!

PR #847, opened 2019-09-27
Feature/improved esim batch management

la3lma:
Progress on support software for managing batches of esims.   Still very preliminary, but the gist of it is that we should progress from a set of "command line scripts" written in golang, to a more coherent program/set of scripts  written in golang using persistent storage somewhere (rdbms, most likely), and from there maybe to a service running somewhere doing these things via a gui.  It will all be driven by actual demand, and no fluff or cosmetics will be permitted into the codebase.

However, it would be nice if PR could also include some changes that will make the build automatically  compile and run through integration tests.  That's not added yet.


sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+
+	state := ParserState{
+		currentState:      INITIAL,
+		inputVariables:    make(map[string]string),
+		headerDescription: make(map[string]string),
+	}
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
```
kmmatwork, line 152: Maybe trim whitespace at start and at end too?

la3lma: Where are they?   Neither Emacs nor I could find any :-)

kmmatwork (approved):
LGTM!

Don't know Go, but could not find any error in the logic as such (easy to follow!). Ensuring that input is stripped from trailing and maybe leading whitespaces might be an idea though?

sim-administration/uploader/outfile_to_hss_input_converter.go
kmmatwork: Was thinking of whitespaces in input...

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+	"strconv"
+	"strings"
+)
+
+type OutputFileRecord struct {
+	Filename          string
+	inputVariables    map[string]string
+	headerDescription map[string]string
+	entries           []SimEntry
+	noOfEntries       int
```
gordonklaus: Is this field unnecessary?  It seems to only duplicate `len(entries)`.

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+			continue
+		} else if isSectionHeader(line) {
+			nextMode := modeFromSectionHeader(line)
+			transitionMode(&state, nextMode)
+			continue
+		}
+
+		// ... or should we look closer at it and parse it
+		// looking for real content?
+		if state.currentState == HEADER_DESCRIPTION {
```
gordonklaus: Consider using `switch state.currentState` instead of if-elseif...

gordonklaus: Switch statements in Go are nice because you don't need to write `break` in every case block.

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+	parsedString := strings.Split(s, " ")
+	return parsedString[0], parsedString[1], parsedString[2]
+}
+
+func transitionMode(state *ParserState, targetState string) {
+	state.currentState = targetState
+}
+
+func modeFromSectionHeader(s string) string {
+	sectionName := s[1:len(s)]
```
gordonklaus: You can just write `s[1:]`; the `len(s)` is implicit.  Same goes for slicing from index zero:  `s[:n]` is equivalent to `s[0:n]`.

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+	if sectionName == "HEADER DESCRIPTION" {
```
gordonklaus: Consider using `switch`.

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+func WriteHssCsvFile(filename string, entries []SimEntry) error {
+
```
gordonklaus, line 336: Consider using https://golang.org/pkg/encoding/csv/?

sim-administration/uploader/outfile_to_hss_input_converter.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	fmt.Println("Successfully written ", max, " sim card records.")
+	return f.Close()
+}
+
+///
+///   Main.
+///
+
+func main() {
```
gordonklaus: I've found it quite common among Go programmers to put `main` at the top of the file, and more generally to list functions as a breadth-first traversal of the call tree (in the absence of recursive calls).  The rationale being that, when you open a file, the first thing you want to see is the highest-level entry point, which is much more relevant than the implementation details of helper functions.

sim-administration/uploader/outfile_to_hss_input_converter_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Check that the output entry set looks legit.
+	assert.Equal(t, 3, len(record.entries))
+	assert.Equal(t, 3, record.noOfEntries)
+}
+
+//
+//  The test suite tying it all together.
+//
+func Test(t *testing.T) {
```
gordonklaus: Remove this function and export (capitalize) the two others.  `go test` will run all the functions starting with `Test` and accepting a `*testing.T`.

sim-administration/uploader/upload-sim-batch.go
```diff
@@ -102,16 +102,16 @@ func isICCID(s string) bool {
 	return match
 }
 
-func checkICCIDSyntax(name string, potentialIccid string) {
+func CheckICCIDSyntax(name string, potentialIccid string) {
```
gordonklaus: There is no need to export (capitalize) these functions.  All the files in a directory belong to the same package and thus have access to all names in that package.

sim-administration/uploader/outfile_to_hss_input_converter_test.go
la3lma: before or after record declarations? (after I presume, but just checking).

sim-administration/uploader/upload-sim-batch.go
la3lma: even when testing?

sim-administration/uploader/outfile_to_hss_input_converter.go
la3lma: ok

sim-administration/uploader/outfile_to_hss_input_converter.go
la3lma: Aaah.  ;-) Thank you (silly me)

sim-administration/uploader/outfile_to_hss_input_converter.go
la3lma: Good idea.  For now I'll add a comment about that, but using that library  is clearly an improvement.

sim-administration/uploader/outfile_to_hss_input_converter.go
la3lma: Don't know yet :-).  The lengt is also a declared length, and the length of the entries should be the same as the declared length, but if there is an error it won't be.  Also I'm not convinced that I should read in the whole set of entries in to memory.   I'll add a TODO about it but leave it for now.

sim-administration/uploader/outfile_to_hss_input_converter_test.go
gordonklaus: I'm not sure what you mean by record declarations (structs?), but the order of top level declarations has no effect on program behavior.

sim-administration/uploader/upload-sim-batch.go
gordonklaus: Yep.  When testing, the package is built "augmented" with the tests, so they belong to the same package as the code.

PR #858, opened 2019-10-02
Remove the no longer used metrics from ocs gateway

mpeterss:
These metrics are no longer in use. Since we are now deploying the OCS Gateway in the cloud, there is no longer any reason to send the analytics events over gRPC to Prime.


vihangpatil (approved):

cyriaquebrousse (approved):
🚿

la3lma:
Is this PR still active? If it is, then I'll comment on it, otherwise it should perhaps be removed?

mpeterss:
it is still active.  Please review !

PR #860, opened 2019-10-03
Refactor activation in the PubSub DataSource

mpeterss:
Making it more usable for MultiDataSource implementation


ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/PubSubDataSource.kt
```diff
@@ -51,29 +51,11 @@ class PubSubDataSource(
             pubSubChannelProvider = FixedTransportChannelProvider.create(GrpcTransportChannel.create(channel))
         }
 
-        // init publisher
-        logger.info("Setting up Publisher for pubsub Topic: {}", ccrTopicId)
         publisher = setupPublisherToTopic(projectId, ccrTopicId)
+        initCcrKeepAlive()
```
vihangpatil: In existing code, the receivers were set before keep-alive.  Does that matter?

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/GrpcDataSource.java
```diff
@@ -252,7 +252,7 @@ private void reconnectStreams() {
         if (!isReconnecting()) {
 
             reconnectStreamFuture = executorService.schedule((Callable<Object>) () -> {
-                        LOG.debug("Reconnecting GRPC streams");
+                        LOG.debug("Reconnecting gRPC streams");
```
vihangpatil, line 255: 👍

vihangpatil (approved):

ocsgw/src/main/java/org/ostelco/ocsgw/datasource/protobuf/PubSubDataSource.kt
mpeterss: Not sure, will fix it

PR #862, opened 2019-10-08
Synchronize client and server schema

havard024:


havard024:
@vihangpatil I think this fixes the issues between client / server schema.

prasanthu (changes requested):
There are more `customer.graphqls` files in this project
```
 ~/dev/ostelco-core ‹develop› » find . | grep customer.graphqls
./graphql/src/test/resources/customer.graphqls
./prime/config/customer.graphqls
```

havard024:
@prasanthu thanks for noticing! I've updated `./graphql/src/test/resources/customer.graphqls` in the previous commit

prasanthu (approved):

PR #866, opened 2019-10-11
Support CCR-I without MSCC

mpeterss:


ocs-ktc/src/main/kotlin/org/ostelco/prime/ocs/core/OnlineCharging.kt
```diff
@@ -58,30 +58,40 @@ object OnlineCharging : OcsAsyncRequestConsumer {
...
-                                            Notifications.lowBalanceAlert(msisdn, consumptionResult.granted, consumptionResult.balance)
-                                            doneSignal.countDown()
-                                        }
-                                )
+                    if (request.msccCount == 0) {
+                        storage.consume(msisdn, 0L, 0L) {
+                            storeResult -> storeResult.fold(
+                                {responseBuilder.resultCode = ResultCode.DIAMETER_USER_UNKNOWN},
+                                {responseBuilder.resultCode = ResultCode.DIAMETER_SUCCESS})
+                        }
```
vihangpatil, line 65: There should be `doneSignal.countDown()` for `if` part too.

mpeterss: I will refactor a bit. The countdownlatch is used in case there are many mscc ( ie many rating-groups). in this case we only fake a mscc so it is not needed here.

vihangpatil (approved):

PR #868, opened 2019-10-11
Feature/improved esim batch management

la3lma:
Refactoring and bug-fixes.    Not done, but now the path towards done-ness seems straightforward.


.gitignore
```diff
@@ -45,3 +45,4 @@ metrics_descriptor.pb
 # Python virtual env
 venv
 prime-service-account.json
+go.sum
```
gordonklaus: Don't ignore go.sum, check it in!  It contains checksums for all your dependencies, so you can be sure _everyone_ is using exactly the same code.

la3lma: Ok. fixing that.

la3lma: ... but apart from that, will you approve the PR? It won't be merged otherwise.

gordonklaus (approved):

PR #871, opened 2019-10-14
On Region Approved Action KTS

vihangpatil:
WARNING: DO NOT MERGE PR.

In prime, we have logic that when a customer is eKYC Approved for a Region, we put that customer in a default segment for that Region - **country-##**.  When we added the membership for some regions, we added one more segment **plan-country-##**. 

These different segments we needed because we wanted to support multiple currencies and tax rates. But we are planning to change that now and have a single currency - **SGD** and single tax rate - that of Singapore.

I’ve moved the logic of static naming of segment **country-##** / **plan-country-##** to _kts-engine_. So, we can change this dynamically without the need for code changes and redeploying prime.

In past, when we moved the logic to dynamically change the free welcome pack, it was done in a generic way using **OnNewCustomerAction**, so that it can do more minor things. For similar reason, the new action is **OnRegionApprovedAction** replacing _getDefaultSegmentForRegion_.


prasanthu (approved):
LGTM

prasanthu (approved):

PR #896, opened 2019-10-24
Address for Malaysia eKYC

vihangpatil:


prime/infra/dev/prime-customer-api.yaml
```diff
@@ -361,11 +347,27 @@ paths:
...
       security:
-        - auth0_jwt: []
         - firebase: []
   "/regions/sg/kyc/profile":
     put:
-      description: "Update Singapore Customer's address and phone number."
+      description: "Update address for Singapore region."
+      produces:
+        - application/json
+      operationId: "updateDetails"
```
prasanthu: Should we use a different operationId ?

prasanthu (approved):
Other than my comment about the operation ID, LGTM

prime/infra/dev/prime-customer-api.yaml
vihangpatil: Yes, I agree.

prasanthu (approved):

PR #914, opened 2019-11-04
Adds handling of recurring payment

kmmatwork:


acceptance-tests/Dockerfile.ngrok
```diff
@@ -0,0 +1,18 @@
+FROM azul/zulu-openjdk:13
```
vihangpatil: Today, I have bumped this version to 13.0.1 which is already merged to develop.

acceptance-tests/src/main/kotlin/org/ostelco/at/TestSuite.kt
```diff
@@ -49,6 +51,41 @@ class TestSuite {
...
+     *    2) Webhooks enabled and configured to use the NGROK endpoint on Stripe.
+     * As NGROK only allows for up to 40 connections per minute, which with Stripe
+     * events means max 40 events per minute, the recurring payment tests has to
+     * be run seperately from the other tests. Otherwise the limit will be exceeded
+     * and tests will fail.
+     * Note that if the NGROK_AUT_TOKEN variable is not set, then the test will be
+     * skipped altogether.
+     */
+    @Test
+    @EnabledIfEnvironmentVariable(named = "NGROK_AUTH_TOKEN", matches="\\S+")
```
vihangpatil: Do you want these tests to be run on CI?  Should we add this ENV var to CI setup?

model/src/main/kotlin/org/ostelco/prime/model/Entities.kt
```diff
@@ -297,26 +291,26 @@ enum class ProductProperties(val s: String) {
 
 enum class ProductClass {
     SIMPLE_DATA,
+    SUBSCRIPTION,
```
vihangpatil: Subscription is not a Product class.  Instead, it is a payment type.
(Product Class = Membership + Payment type = Subscription) => Our Membership project which we have removed.
(Product Class = Simple Data + Payment type = Subscription) => Product which will topup `x` bytes evey `y` period.
(Product Class = Data with Validity (coming soon) + Payment type = Subscription) => We will have this.

neo4j-store/src/main/resources/AcceptanceTestSetup.kts
```diff
@@ -126,4 +129,71 @@ adminStore.atomicCreateOffer(
...
+        plan = Plan(
+                id = "PLAN_10USD_DAY",
+                interval = "day",
+                trialPeriod = 4000L),      // 4 sec. trial time
+        stripeProductName = "Daily subscription plan",
+        planProduct = Product(
+                sku = "PLAN_10USD_DAY",
+                price = Price(amount = 10_00, currency = "USD"),
+                properties = mapOf(
+                        PRODUCT_CLASS.s to SUBSCRIPTION.name,
```
vihangpatil: Similar to another comment, product class should be MEMBERSHIP.
Payment type is SUBSCRIPTION.
If no payment type is set, then the default value is ONE TIME PAYMENT.

neo4j-store/src/main/resources/AcceptanceTestSetup.kts
```diff
@@ -126,4 +129,71 @@ adminStore.atomicCreateOffer(
...
+                        SEGMENT_IDS.s to "country-us"),
+                payment = mapOf(
```
vihangpatil, line 164: The payment type = subscription has to be set in payment properties.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/Products.kt
```diff
@@ -32,6 +32,24 @@ val expectedPlanProduct: Product = Product()
...
+                        "segmentIds" to "country-us"
+                )
+        )
+        .payment(
+                mapOf(
+                        "type" to "SUBSCRIPTION",
+                        "label" to "Daily subscription plan",
+                        "taxRegionId" to "us"
+                )
+        )
```
vihangpatil: Here, the payment type is set correctly to "SUBSCRIPTION", but the product class should be MEMBERSHIP.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
```diff
@@ -54,4 +55,31 @@ class ProductsResource(private val dao: SubscriberDAO) {
                         saveCard = saveCard ?: false)
                         .responseBuilder(Response.Status.CREATED)
             }.build()
+
+    @POST
+    @Path("{sku}/renew")
+    @Produces(MediaType.APPLICATION_JSON)
+    fun renewSubscription(@Auth token: AccessTokenPrincipal?,
```
vihangpatil: Just the word `Subscription` means mobile Subscription (with MSISDN as ID).  Better to call this `PaymentSubscription` or `PlanSubscription` to distinguish it from "Mobile" Subscription.

app-notifier/src/main/kotlin/org/ostelco/prime/appnotifier/FirebaseAppNotifier.kt
```diff
@@ -45,6 +45,24 @@ class FirebaseAppNotifier: AppNotifier {
...
+                    /* TODO: Add support for 3D secure notification. */
+                }
+                NotificationType.SUBSCRIPTION_RENEWAL_UPCOMING -> {
+
+                }
+                NotificationType.SUBSCRIPTION_RENEWAL_STARTING -> {
+                    /* No notification, as a notification will be sent on either successful
+                       or failed renewal. */
+                }
             }
```
vihangpatil, line 66: This API is a bit messed up.  `enum NotificationType` can have fields such as `title` and `body` which is then directly used instead of using `when` cases.  Also, if FCMStrings is specific to Firebase, they should not be in generic API interfaces - `prime-modules`.

The simplest way to design it is to think of `AppNotifier` as an interface whereas `FirebaseAppNotifier` as one of the many possible implementations of `AppNotifier`.
We should be able to replace whole FCM with some other implementation with just adding and removing dependency in the gradle build file and no need of any code change.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -458,10 +513,34 @@ class StripePaymentProcessor : PaymentProcessor {
...
+     */
+    override fun payInvoice(invoiceId: String, sourceId: String?): Either<PaymentError, InvoicePaymentInfo> =
             either("Failed to complete payment of invoice ${invoiceId}") {
-                val receipt = Invoice.retrieve(invoiceId).pay()
-                InvoicePaymentInfo(receipt.id, receipt?.charge ?: UUID.randomUUID().toString())
+                val params = mapOf(
+                        *(if (sourceId != null)
+                            arrayOf("source" to sourceId)
+                        else arrayOf())
+                )
```
vihangpatil: Outer if-else with mapOf inside is more readable.
Also, ?. is more compact way to check nulls.

acceptance-tests/src/main/kotlin/org/ostelco/at/TestSuite.kt
kmmatwork: Yes, but we will first have to create an account for CI with ngrok.io first. And maybe consider a baslc or pro subscription? Being able to use "custom subdomain" would simplify the setup a bit...

app-notifier/src/main/kotlin/org/ostelco/prime/appnotifier/FirebaseAppNotifier.kt
kmmatwork: https://www.notion.so/redotter/Review-FCM-delivery-setup-reliability-and-message-format-0701c7da03e24a89963cc7bdd4f7aa76

model/src/main/kotlin/org/ostelco/prime/model/Entities.kt
kmmatwork: Maybe it could be an idea to rewrite the "payment paths" in the code to only consider the `payment[type]` field and not consider the `productClass` field at all? And maybe make the `payment[type]` field mandatory?

kmmatwork: Don't know yet how straight forward this will be though. Will have to check first...

kmmatwork:
Fyi, more changes to how plans/products gets stored will follow later as part of the 
https://www.notion.so/redotter/Document-how-Backend-should-handle-prorate-price-plan-changes-for-a-given-Product-39d34912997c4a74870b1960c9fa08cb
and the 
https://www.notion.so/redotter/Change-implementation-of-Purchase-Records-to-store-Price-information-from-Stripe-Invoices-f3ac8e7f00c84cc1b60d01ef5512712a
tasks.

acceptance-tests/script/wait.sh
```diff
@@ -220,7 +220,18 @@ EOF
 echo " ... HLR preactivated rofiles loaded into prime"
 
 
+##
+## Creating pub-sub subscriptions for recurring payments tests (plans).
+## Have to create one subscription per test, as the tests runs in parallel.
+##
+
+echo "Creating pub-sub subscriptions for recurring payment tests"
+curl -X PUT -H "Content-Type: application/json" -d '{"topic":"projects/'${GCP_PROJECT_ID}'/topics/stripe-event","ackDeadlineSeconds":10}' pubsub-emulator:8085/v1/projects/${GCP_PROJECT_ID}/subscriptions/stripe-event-okhttp-purchase-ok-sub
```
la3lma, line 229: This is so much repeated code that it now arguably would make sense to factor this out into a function.   Or not. Your call.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+                                            logger.info("Received 'invoice.payment_succeeded' event for customer $customerId")
+                                    }
+                        }
+                    })
+
+    fun waitForFailedSubscriptionRenewal(topic: String = StripeEventListener.topic,
+                                         subscription: String,
+                                         customerId: String,
+                                         timeout: Long = 10000L): Boolean =
+            waitForStripeEvent(
```
la3lma, line 52: This code  has very convoluted control flow. Consider breaking it up into smaller pieces with descriptive names.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+
+        try {
+            runBlocking {
+                status = withTimeoutOrNull(timeout) {
+                    subscriber.start()
+                    while (!subscriber.eventSeen()) {
+                        delay(200L)
+                    }
+                    "Done"
+                } != null
```
la3lma, line 86: This makes it very obscure what we're checking for.   Please consider making a local function with a descriptive name.  Code should preferably be trivial to read, and this isn't.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -66,6 +69,29 @@ object StripePayment {
...
+                        "exp_year" to nextYear(),
+                        "cvc" to "314"),
+                "owner" to mapOf(
+                        "address" to mapOf(
+                                "city" to "Oslo",
+                                "country" to "Norway"
+                        ),
+                        "email" to "me@somewhere.com")
+        )
+        val source = Source.create(sourceMap)
```
la3lma, line 91: I would have preferred calling this paymentSource, just to make it blindingly obvious to casual readers (such as myself, in the role of reviewer) what this value is.

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -1897,7 +1899,157 @@ class PlanTest {
...
+
+            val sourceId = StripePayment.createPaymentTokenId()
+
+            post<String> {
+                path = "/products/$sku/purchase"
+                this.email = email
+                queryParams = mapOf("sourceId" to sourceId)
+            }
+
+            Thread.sleep(200) // wait for 200 ms for balance to be updated in db
```
la3lma, line 1976: This is brittle.   Is there a less brittle way to wait for the database?

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -1897,7 +1899,157 @@ class PlanTest {
...
+            // First record - free product
+            // Second record - product subscribed to
+            assert(2 == purchaseRecords.size) { "Got ${purchaseRecords.size} purchase records, expected 2" }
+            assertEquals(expectedPlanProductUS, purchaseRecords.last().product, "Incorrect 'Product' in purchase record")
+
+            // Actual charge for renewal will first be done after trial time
+            // expires, which will happen after 4 sec. Waiting a bit longer
+            // before checking the outcome.
+            StripeEventListener.waitForSubscriptionPaymentToSucceed(
+                    subscription = "stripe-event-jersey-purchase-ok-sub",
```
la3lma, line 1995: What does this magic string mean? Consider putting it into a variable with a descriptive name.

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+
+        val email = "purchase-${randomInt()}@test.com"
+        val sku = "PLAN_10USD_DAY"
+        var customerId = ""
+
+        try {
+            customerId = createCustomer(name = "Test Purchase Plan User", email = email).id
+            enableRegion(email = email, region = "us")
+
+            val client = clientForSubject(subject = email)
```
la3lma, line 1066: There is a lot of repeated code between this and the previous test.   Consider factoring that code into separate methods or perhaps even running the code in a a "setUp" method as part off fixture setup for the test suite.

Repeated code obscures the intent and implementation of the actual test.

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+
+            // Switch to a "no funds" card.
+            // This should cause subscription renewal to fail.
+            client.createSource(insuffientFundsSourceId)
+            client.setDefaultSource(insuffientFundsSourceId)
+
+            // Actual charge for renewal will first be done after trial time
+            // expires, which will happen after 4 sec. Waiting a bit longer
+            // before checking the outcome.
+            StripeEventListener.waitForFailedSubscriptionRenewal(
```
la3lma, line 1093: Repeated code

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+            // (200 ms is too low...)
+            Thread.sleep(2000)
+
+            purchaseRecords = client.purchaseHistory
+            assert(2 == purchaseRecords.size) { "Got ${purchaseRecords.size} purchase records, expected 2 due to renewal failure" }
+
+            // Explicitly renew the subscription with a new card.
+            client.renewSubscription(sku, newSourceId, true)
+
+            StripeEventListener.waitForSubscriptionPaymentToSucceed(
```
la3lma, line 1108: Repeated code

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
```diff
@@ -54,4 +54,31 @@ class ProductsResource(private val dao: SubscriberDAO) {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    fun renewPaymentSubscription(@Auth token: AccessTokenPrincipal?,
+                                 @NotNull
+                                 @PathParam("sku")
+                                 sku: String,
+                                 @QueryParam("sourceId")
+                                 sourceId: String?,
+                                 @QueryParam("saveCard")
+                                 saveCard: Boolean?): Response =
+            if (token == null) {
```
la3lma, line 69: Can authorisation check be delegated to a filter so that it does not show up in the business logic part of the resource implementation?  Technically it obviously can, but the question is obviously if it makes sense in this particular situation.  If  the "token == null"  must be there, add a comment describing why this couldn't be delegated to a filter, of if it can, at the very least add a "TODO" indicating that it should.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
```diff
@@ -54,4 +54,31 @@ class ProductsResource(private val dao: SubscriberDAO) {
...
+                Response.status(Response.Status.UNAUTHORIZED)
+            } else {
+                if (sourceId == null) {
+                    dao.renewPaymentSubscription(
+                            identity = token.identity,
+                            sku = sku)
```
la3lma, line 75: In this case the saveCard parameter is discarded?   If it's non-null that should  indicate an error, or it should be used.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
```diff
@@ -54,4 +54,31 @@ class ProductsResource(private val dao: SubscriberDAO) {
...
+                } else {
+                    dao.renewPaymentSubscription(
+                            identity = token.identity,
+                            sku = sku,
+                            sourceId = sourceId,
```
la3lma, line 80: What is the default value of sourceId? If it's null, then perhaps only the "else" branch is necessary and the code can be made more compact without losing generality?

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        try {
+            runBlocking {
+                status = withTimeoutOrNull(timeout) {
+                    subscriber.start()
+                    while (!subscriber.eventSeen()) {
+                        delay(200L)
+                    }
+                    "Done"
```
la3lma, line 85: This control flow is very obscure. Can it be made more obvious?

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+                subscription = subscription,
+                cmp = cmp
+        )
+
+        try {
+            runBlocking {
+                status = withTimeoutOrNull(timeout) {
+                    subscriber.start()
+                    while (!subscriber.eventSeen()) {
+                        delay(200L)
```
la3lma, line 83: Why the delay, why 200, which unit is it (milliseconds).  Please explain, at least in a comment.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+                                         subscription: String,
+                                         customerId: String,
+                                         timeout: Long = 10000L): Boolean =
+            waitForStripeEvent(
+                    topic = topic,
+                    subscription = subscription,
+                    eventType = "customer.subscription.updated",
+                    timeout = timeout,
+                    cmp = {
+                        if (it.type != "customer.subscription.updated")
```
la3lma, line 58: Obscure controll flow. Make it more obvious what's going on.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+            logger.error("ACCEPTANCE TEST PUBSUB: Error disconnecting to service: {}",
+                    e.message)
+        }
+    }
+
+    private fun handler(message: ByteString, consumer: AckReplyConsumer) {
+        try {
+            gson.fromJson(message.toStringUtf8(), Event::class.java)
+                    .let {
+                        if (cmp(it)) eventSeen = true
```
la3lma, line 152: What does this mean? My initial reading is  "compare it with itself", which makes no sense.  A comment is in order.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -105,6 +131,7 @@ object StripePayment {
      * Obtains the Stripe 'customerId' directly from Stripe.
      */
     fun getStripeCustomerId(customerId: String) : String {
+
         // https://stripe.com/docs/api/java#create_card_token
         Stripe.apiKey = System.getenv("STRIPE_API_KEY")
```
la3lma, line 136: This thing doesn't test if the STRIPE_API_KEY is non-null. It's clearly an error if it is, and that error should be detected and signalled.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -121,16 +148,21 @@ object StripePayment {
     }
 
     fun deleteCustomer(customerId: String) {
+
         // https://stripe.com/docs/api/java#create_card_token
         Stripe.apiKey = System.getenv("STRIPE_API_KEY")
```
la3lma, line 153: Getting the stripe API key should be wrapped in a method that checks if it has a valid value (at least that it isn't empty), and throws/signals the appropriate error if it's not valid.   The rationale behind this request is: This will fail eventually, so it is prudent to fail as early as possible to make sure that half-committed values are not stored anywhere before the error is detected.

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -1897,7 +1899,157 @@ class PlanTest {
...
+                path = "/paymentSources"
+                this.email = email
+                queryParams = mapOf("sourceId" to insuffientFundsSourceId)
+            }
+
+            // Actual charge for renewal will first be done after trial time
+            // expires, which will happen after 4 sec. Waiting a bit longer
+            // before checking the outcome.
+            StripeEventListener.waitForFailedSubscriptionRenewal(
+                    subscription = "stripe-event-jersey-purchase-fail-sub",
```
la3lma, line 2065: This (along with other things) in this test is repeated code. Repeated code makes the code less readable in the sense that it's difficult to know what's special in something that looks similar but can't be known to be similar before a complete inspection has been done.  It is better to factor this out into a separate variable or method (depending on the context) so that it is obvious, at a glance, what is identical and what isn't, and also what the differences are (can be highlighted using parameters to methods)).

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+
+            // First record - free product
+            // Second record - product subscribed to
+            assert(2 == purchaseRecords.size) { "Got ${purchaseRecords.size} purchase records, expected 2" }
+            assertEquals(expectedPlanProductUS, purchaseRecords.last().product, "Incorrect 'Product' in purchase record")
+
+            // Actual charge for renewal will first be done after trial time
+            // expires, which will happen after 4 sec. Waiting a bit longer
+            // before checking the outcome.
+            StripeEventListener.waitForSubscriptionPaymentToSucceed(
```
la3lma, line 1040: Repeated code

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+                    subscription = "stripe-event-okhttp-purchase-ok-sub",
+                    customerId = customerId,
+                    timeout = 30000L)
+
+            // Wait for DB to be updated.
+            Thread.sleep(200)
```
la3lma, line 1046: Sleeping is a very brittle way to ensure that a database is updated. Is there a better way?

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
```diff
@@ -994,7 +996,125 @@ class PlanTest {
...
+            // Switch to a "no funds" card.
+            // This should cause subscription renewal to fail.
+            client.createSource(insuffientFundsSourceId)
+            client.setDefaultSource(insuffientFundsSourceId)
+
+            // Actual charge for renewal will first be done after trial time
+            // expires, which will happen after 4 sec. Waiting a bit longer
+            // before checking the outcome.
+            StripeEventListener.waitForFailedSubscriptionRenewal(
+                    subscription = "stripe-event-okhttp-purchase-fail-sub",
```
la3lma, line 1094: Repeated code

model/src/main/kotlin/org/ostelco/prime/model/Entities.kt
```diff
@@ -332,6 +337,9 @@ data class RefundRecord(
         val reason: String, // possible values are duplicate, fraudulent, and requested_by_customer
         val timestamp: Long) : HasId
 
+/* TODO! (kmm) To support prorating with subscriptions, the
```
la3lma, line 344: what is "prorating"?

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -1163,18 +1173,29 @@ object Neo4jStoreSingleton : GraphStore {
                         .bind()
 
                 if (product.price.amount > 0) {
-                    val (chargeId, invoiceId) = when (product.paymentType) {
-                        SUBSCRIPTION -> {
-                            val subscriptionDetailsInfo = purchasePlan(
+                    val purchaseRecord = when (product.paymentType) {
```
la3lma: This is a very long if-branch.  Consider making a method with a descriptive name (a scope-local one if convenient). This will increase the readability of the code.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -2437,45 +2468,221 @@ object Neo4jStoreSingleton : GraphStore {
...
+                                getProduct(identity, sku)   /* Return the subscribed to product. */
+                            }
+                }
+    }
+
+    override fun renewSubscriptionToPlan(identity: org.ostelco.prime.model.Identity,
+                                         sku: String,
+                                         sourceId: String,
+                                         saveCard: Boolean): Either<StoreError, Product> = writeTransaction {
+        getCustomer(identity = identity)
```
la3lma, line 2597: This is highly unreadable code.  The control flow is obscure, the level of indentation is high.  Please introduce other abstractions (local methods probably) to make the control flow and thus semantics of this statement comprehensible.  Right now it isn't.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -2437,45 +2468,221 @@ object Neo4jStoreSingleton : GraphStore {
...
+                    Unit.right()
+                }
+    }
+
+    override fun notifySubscriptionToPlanRenewalStarting(customerId: String,
+                                                         sku: String,
+                                                         invoiceId: String,
+                                                         payInvoiceNow: Boolean): Either<StoreError, Unit> = readTransaction {
+        get(Product withSku sku)
+                .flatMap { product ->
```
la3lma, line 2670: Obscure control flow. See comment above about similar issue.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -2437,45 +2468,221 @@ object Neo4jStoreSingleton : GraphStore {
...
-                            .bind()
-                }
-                logger.info("Customer $customerId completed payment of invoice $invoiceId for subscription to plan ${plan.id}")
-
+                logger.info(
+                        "Customer $customerId completed payment of invoice ${subscriptionPaymentInfo.invoiceId} " +
+                        "for renewal of subscription to plan ${plan.id}"
+                )
                 plan
             }.fix()
```
la3lma: This boilerplate adds two levels of indentation and obscures control flow.  Can it be abstracted away into something more compact?

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -2437,45 +2468,221 @@ object Neo4jStoreSingleton : GraphStore {
...
+                /* Notify the customer about pending renewal. */
+                appNotifier.notify(
+                        notificationType = NotificationType.PAYMENT_METHOD_REQUIRED,
+                        customerId = customerId,
+                        data = mapOf(
+                                "sku" to sku
+                        ).plus(product.presentation))
+                Unit
+            }.fix()
+        }.unsafeRunSync()
```
la3lma: Same as above, can some abstraction with a lambda ba used instead, e.g. "withinTransaction(transaction) { ...}" or something?

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -196,6 +209,35 @@ class StripePaymentProcessor : PaymentProcessor {
                 ProductInfo(product.delete().id)
             }
 
+    override fun addSource(customerId: String,
+                           sourceId: String,
+                           setDefault: Boolean): Either<PaymentError, SourceInfo> =
+            either("Failed to add source $sourceId to customer $customerId") {
```
la3lma, line 227: Very obscure control flow.  Please rewrite for readability (whole method)

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -223,63 +265,75 @@ class StripePaymentProcessor : PaymentProcessor {
...
                                 { it }
                         )
-                val subscriptionParams = mapOf(
+                val trialEndInEpochSeconds = ofEpochMilliToSecond(trialEnd)
+                val params = mapOf(
                         "customer" to stripeCustomerId,
                         "items" to mapOf("0" to item),
-                        *(if (trialEnd > Instant.now().epochSecond)
-                            arrayOf("trial_end" to trialEnd.toString())
+                        *(if (trialEndInEpochSeconds > Instant.now().epochSecond)
```
la3lma, line 293: What does the asterisk mean in this context? This is necessarily obscure.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -223,63 +265,75 @@ class StripePaymentProcessor : PaymentProcessor {
...
+        return if (subscription.status == "trialing") {
+            PaymentStatus.TRIAL_START
+        } else {
+            require(intent != null) {
+                "Expected an invoice intent with invoice ${invoice.id} for subscription ${subscription.id} " +
+                        "with status ${subscription.status}"
+            }
+            when (subscription.status) {
+                "active" -> {
+                    when (intent.status) {
```
la3lma, line 318: Please avoid using magic strings like "succeed", "incomplete" etc. If these refer to protocol options, then please use enumerations with strings in them to identify the different protocol options.   Naked (and thus magic) strings obscure what's going on, and are also difficult to keep track of if used more than one place in the codebase (and it's hard to know if it's used more than once since text search is necessary, it's much easier to  jump to the enumeration and then see where it's used).

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -223,63 +265,75 @@ class StripePaymentProcessor : PaymentProcessor {
...
-                Triple(PaymentStatus.TRIAL_START, "", "")
-            }
-            else -> {
-                throw RuntimeException(
-                        "Got unexpected status ${subscription.status} when creating subscription ${subscription.id}")
-            }
+        }.let {
+            SubscriptionPaymentInfo(id = subscription.id,
+                    status = it,
+                    invoiceId = if (invoice != null)
```
la3lma, line 340: consider using either braces in branches of if/else or wrapping it in a () (or both).

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripeUtils.kt
```diff
@@ -23,29 +24,33 @@ object StripeUtils {
         return try {
             Either.right(action())
         } catch (e: CardException) {
-            // If something is decline with a card purchase, CardException will be caught
-            logger.warn("Payment error : $errorDescription , Stripe Error Code: ${e.code}")
-            Either.left(ForbiddenError(errorDescription, e.message))
+            /* Something is wrong with the payment source/card.
+               402 indicates that the card is out of funds or expired etc. */
+            logger.warn("Payment error: ${errorDescription}, Stripe Error Code: ${e.code}")
+            Either.left(if (e.statusCode == 402)
```
la3lma: Please don't use naked numbers.   Find a constant that represents the semantics of this particular return code (I didn't remember  off hand that 402 means "payment required", something that would have been obvious if a descriptive enumeration name had been used instead). There are plenty of them in the various HTTP utility classes we use, please use one of them.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripeUtils.kt
```diff
@@ -23,29 +24,33 @@ object StripeUtils {
...
+                PaymentFailedError(errorDescription, e.message)
+            else
+                ForbiddenError(errorDescription, e.message))
```
la3lma: What is forbidden error? An error because something is forbidden, or an error that should never happen?   If it's the latter then the method should be given a better name,, otherwise some comment should indicate why the invocation was forbidden.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/subscribers/RecurringPaymentStripeEvent.kt
```diff
@@ -49,57 +58,146 @@ class RecurringPaymentStripeEvent : PubSubSubscriber(
                     }
             )
 
+    /* Upcoming renewal and successfully renewed subscriptions. */
     private fun invoiceEvent(eventType: String, invoice: Invoice) {
-        /* Skip invoices not related to subscriptions (recurring payment). */
         if (invoice.subscription.isNullOrEmpty())
             return
+        if (invoice.billingReason != "subscription_cycle")
```
la3lma, line 65: Magic string! Use string constant or string associated with enumeration instead.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/subscribers/RecurringPaymentStripeEvent.kt
```diff
@@ -49,57 +58,146 @@ class RecurringPaymentStripeEvent : PubSubSubscriber(
...
-                    invoice.lines.data.forEach {
-                        purchasedSubscriptionEvent(invoice.customer, invoice.id, invoice.charge, it.plan)
-                    }
-                else
-                    logger.debug("Invoice ${invoice.id} successfully paid with billing reason: ${invoice.billingReason}")
-            }
-            "invoice.payment_failed" -> {
-            }
+            /* Early notification (configurable in Stripe console) of renewal. */
             "invoice.upcoming" -> {
```
la3lma, line 70: More magic strings.

prime-modules/src/main/kotlin/org/ostelco/prime/apierror/ApiErrorCodes.kt
```diff
@@ -65,6 +65,7 @@ enum class ApiErrorCode {
     FAILED_TO_STORE_PLAN,
     FAILED_TO_REMOVE_PLAN,
     ALREADY_SUBSCRIBED_TO_PLAN,
+    FAILED_TO_RENEW_SUBSCRIPTION,
     // FAILED_TO_SUBSCRIBE_TO_PLAN,
```
la3lma, line 70: If this shouldn't be enabled in the immediate  future, then remove the dead code.

prime-modules/src/main/kotlin/org/ostelco/prime/paymentprocessor/core/Model.kt
```diff
@@ -18,18 +18,37 @@ data class ProfileInfo(val id: String)
...
-data class SubscriptionDetailsInfo(val id: String, val status: PaymentStatus, val invoiceId: String, val chargeId: String, val created: Long, val trialEnd: Long = 0L)
-
-data class TaxRateInfo(val id: String, val percentage: BigDecimal, val displayName: String, val inclusive: Boolean)
+data class SubscriptionPaymentInfo(val id: String,                  /* Stripe subscription id. */
+                                   val sku: String,
+                                   val status: PaymentStatus,
+                                   val invoiceId: String,           /* Not set on plans with trial time. */
+                                   val chargeId: String? = null,    /* Only set on successful payment on creation/renewal. */
+                                   val created: Long,
+                                   val currentPeriodStart: Long,
```
la3lma, line 33: Add a comment indicating what the times should be interpreted as? Is it seconds since epoch? If so, say so.

prime/infra/dev/prime-customer-api.yaml
```diff
@@ -651,6 +651,39 @@ paths:
...
+          in: query
+          description: "Whether to save this card as a source for this user (default = false)"
+          required: false
+          type: boolean
+      responses:
+        201:
+          description: "Successfully renewed the subscription"
+          schema:
+            $ref: '#/definitions/Product'
+        404:
```
la3lma, line 706: Wasn't there a 402 in there somewhere also?

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
```diff
@@ -458,10 +513,33 @@ class StripePaymentProcessor : PaymentProcessor {
...
+            retrieved first.
+
+       TODO Must have two pay methods, one for subscription renewal and one for all other
+            or maybe not...
+     */
+    override fun payInvoice(invoiceId: String, sourceId: String?): Either<PaymentError, InvoicePaymentInfo> =
             either("Failed to complete payment of invoice ${invoiceId}") {
-                val receipt = Invoice.retrieve(invoiceId).pay()
-                InvoicePaymentInfo(receipt.id, receipt?.charge ?: UUID.randomUUID().toString())
+                val params = if (sourceId != null)
```
la3lma, line 551: Please wrap if in parens.  I get confused by this kind of implicitly bounded values.  Be explicit.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+        var status = false
+        val subscriber = PubSubSubscriber(
+                topic = topic,
+                subscription = subscription,
+                cmp = cmp
+        )
+
+        try {
+            runBlocking {
+                status = withTimeoutOrNull(timeout) {
```
la3lma, line 80: If this is just a busy-wait loop, then say so in a comment. If not then explain what's going on.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
```diff
@@ -0,0 +1,167 @@
...
+                            val event = it.data.`object` as Invoice
+                            (event.customer == customerId)
+                                    .also {
+                                        if (it)
+                                            logger.info("Received 'invoice.payment_succeeded' event for customer $customerId")
+                                    }
+                        }
+                    })
+
+    fun waitForFailedSubscriptionRenewal(topic: String = StripeEventListener.topic,
```
la3lma, line 48: The control flow of this method is very obscure. Please make it simpler to read by breaking it up into smaller methods with descriptive names rather than one big open-coded method with a  very complex internal life.

la3lma (changes requested):

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
kmmatwork: Kotlins [spread](https://kotlinlang.org/docs/reference/functions.html#variable-number-of-arguments-varargs) operator.

model/src/main/kotlin/org/ostelco/prime/model/Entities.kt
kmmatwork: Standard term for ensuring correct charges when changing/terminating subscriptions etc. More info at [https://stripe.com/docs/billing/subscriptions/prorations](https://stripe.com/docs/billing/subscriptions/prorations) for how to use it with Stripe. Also described many other places.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
kmmatwork: Agree with this. I think the same pattern is used in all client facing APIs, and an update should probably fix it for all of them. Will leave a TODO and a Notion tech dept task for fixing this.

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
kmmatwork: Not sure. The same `wait` statement is present in all acceptance tests that involves updates to `neo4j`. Will add a TODO and a Notion tech dept task.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
kmmatwork: Have to disagree here, provided that you mean the `either()` call. The `either` wrapper i a `try catch` wrapper that will turn exception thrown on calls to Stripe into an `Either`. Used in all methods calling Stripe.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
kmmatwork: Will try to move the `cmp` function out out.

acceptance-tests/script/wait.sh
kmmatwork: Will try to move the `curl` call into a function.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripeEventListener.kt
kmmatwork: Basically cut and paste from [https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html](https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html) at the bottom of the page.

acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
kmmatwork: Good point!

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
kmmatwork: Name of a  subscription to the "stripe-event" pub-sub topic. We need one such subscription for each "recurring payment" acceptance test. The  `waitForSubscriptionPaymentToSucceed` method is there to ensure that we don't wait too long for the expected events.

Not sure if it make sense use a variable for the subscription name. But some more comments about what is going on could be an idea.

acceptance-tests/src/main/kotlin/org/ostelco/at/okhttp/Tests.kt
kmmatwork: Will try to restructure a bit.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
kmmatwork: Will fix.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/ProductsResource.kt
kmmatwork: Yes, this API needs some rethinking.

prime-modules/src/main/kotlin/org/ostelco/prime/paymentprocessor/core/Model.kt
kmmatwork: Internally timestamps are always in milliseconds since epoch.

payment-processor/src/main/kotlin/org/ostelco/prime/paymentprocessor/StripePaymentProcessor.kt
kmmatwork: Agree with that.

PR #924, opened 2019-11-06
Disable notification when user cancels KYC 

vihangpatil:
Do not notify the app when the verification fails because the customer has cancelled and hence did not upload an ID.


sgoodwin (approved):
<3

prasanthu (approved):

PR #930, opened 2019-11-11
Moving shared classes from kts

mpeterss:


mpeterss:
Can only be merged in sync with changes of KTS scripts in dev and prod

vihangpatil (approved):

PR #933, opened 2019-11-12
Use parameters/props in cypher create/set queries

prasanthu:


vihangpatil:
@prasanthu 
The original code sets every property value in string.  Does the new logic change that?  Will it break for other classes which have non-string fields.

kmmatwork (approved):
OK from me!

prasanthu:
> @prasanthu
> The original code sets every property value in string. Does the new logic change that? Will it break for other classes which have non-string fields.

Looking into this.

vihangpatil (approved):

PR #940, opened 2019-11-14
[WIP] Use coroutineScope and async instead of CountDownLatch

mpeterss:
Using deferred to wait for all async calls to end instead of using
a CountDownLatch.


prasanthu (commented):
Try This instead 
[p1.txt](https://github.com/ostelco/ostelco-core/files/3846875/p1.txt)

prasanthu (approved):
Do as we discussed

mpeterss:
Do not merge this, I want to test this more to be sure it is working correctly

vihangpatil (changes requested):
DO NOT MERGE THIS.  The Kotlin Coroutine might not work.

The scope will work when the child coroutines are async using suspend.
In that case, the callback is handled as a continuation.

But in this case, the callback is an execution of the lambda function which is passed as the last parameter.

So, the parent coroutine scope will be blocked till all the coroutines in it are completed, but they get completed immediately.  The scope will not be blocked until all the callbacks are executed.

prasanthu:
> DO NOT MERGE THIS. The Kotlin Coroutine might not work.
> 
> The scope will work when the child coroutines are async using suspend.
> In that case, the callback is handled as a continuation.
> 
> But in this case, the callback is an execution of the lambda function which is passed as the last parameter.
> 
> So, the parent coroutine scope will be blocked till all the coroutines in it are completed, but they get completed immediately. The scope will not be blocked until all the callbacks are executed.

Are you sure ?, @mpeterss Checked this using AT.

prasanthu:
> DO NOT MERGE THIS. The Kotlin Coroutine might not work.
> 
> The scope will work when the child coroutines are async using suspend.
> In that case, the callback is handled as a continuation.
> 
> But in this case, the callback is an execution of the lambda function which is passed as the last parameter.
> 
> So, the parent coroutine scope will be blocked till all the coroutines in it are completed, but they get completed immediately. The scope will not be blocked until all the callbacks are executed.

Are you talking about the bimap lamda? I don't see any asynchronous calls than the `async` in  `consumeRequestHandler` . If the whole loop (`request.msccList.forEach`) runs synchronously, the asyncs methods will be called before the loop is finished. In such cases the async calls are part of the `coroutineScope`

ocs-ktc/src/main/kotlin/org/ostelco/prime/ocs/core/OnlineCharging.kt
```diff
@@ -96,57 +94,56 @@ object OnlineCharging : OcsAsyncRequestConsumer {
...
-                                { consumptionRequest ->  consumeRequestHandler(consumptionRequest) }
-                        )
-            } else {
-                doneSignal.countDown()
+                suspend fun consumeRequestHandler(consumptionRequest: ConsumptionRequest) {
+                    async {
+                        storage.consume(
+                                msisdn = consumptionRequest.msisdn,
+                                usedBytes = consumptionRequest.usedBytes,
+                                requestedBytes = consumptionRequest.requestedBytes) { storeResult ->
```
vihangpatil, line 126: I don't think that the coroutine scope will wait til the last argument to `storage.consume` - the lambda with storeResult arg is called back.

mpeterss: correct. I have seen it fail in AT. So this is not working

prasanthu: Now I understand what you are talking about

prasanthu: `storage.consume()` calls it callback in a different `CoroutineScope`. A new one is created when we call `writeSuspended()`

prasanthu (commented):
Review my changes

prasanthu:
I have reverted the change I made for converting CompletionStage to Deferred. The load test with `10_000` requests failed with this change. The max I could do was `500`

prasanthu:
That revert made the `creditControlRequestInitTerminateNoCredit` fail. So bringing the change back. We need to sit together and understand this

PR #942, opened 2019-11-15
In context, return SimProfiles which are missing in the SIM Manager DB.

vihangpatil:


neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+                                eSimActivationCode = "Dummy eSIM",
+                                msisdnList = emptyList()
+                        )
+                    }
+
+                    org.ostelco.prime.model.SimProfile(
+                            iccId = simProfile.iccId,
+                            alias = simProfile.alias,
+                            eSimActivationCode = simEntry.eSimActivationCode,
+                            status = simEntry.status)
```
kmmatwork, line 843: Not so sure about `status`. The SIM DB has the field 'provisionstate' which keep track of the state of the SIM (AVAILABLE, RESERVED, PROVISIONED, ALLOCATION_FAILED) and `simManager.getSimProfile()` ignores  this value I think. Depends on context whether this is OK or not...

kmmatwork: A SIM entry in the DB has three state fields:
   - hlrstate
   - smdplusstate
   - provisonstate
and all of them must be taken into account if you want the actual state of  a SIM entry.

kmmatwork: But no problem with the change as such.

kmmatwork (approved):
Some nitpicking about the `status` field which I kind of  suspect that can be set to an incorrect value as it don't takes all SIM entry state fields into account when determining the value.

tools/prime-admin/src/main/kotlin/org/ostelco/tools/prime/admin/actions/Utils.kt
```diff
@@ -25,7 +25,7 @@ fun <L, R> Either<L, R>.print() = this.fold(
             this
         },
         { right ->
```
la3lma, line 27: what is the right. thing here? Please use a variable  name that is a bit more descriptive

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
-                                    .mapLeft { NotFoundError(type = simProfileEntity.name, id = simProfile.iccId) }
-                                    .bind()
-                            org.ostelco.prime.model.SimProfile(
-                                    iccId = simProfile.iccId,
-                                    alias = simProfile.alias,
-                                    eSimActivationCode = simEntry.eSimActivationCode,
-                                    status = simEntry.status)
-                        }
-                    }.fix()
+            Either.monad<StoreError>().binding {
```
la3lma, line 809: It is. very difficult. to grasp what goes on.   It would be very good if it could ber broken up into smaller pieces/functions  with clear intent stated.  As it is both he intent and the control flow is very obscure.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+                simProfiles.bind().map { simProfile ->
```
la3lma, line 810: Which map is this? Is it the arrow one, or the kotlin one,?

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+
+                    val regionId = (regionCode ?: map[simProfile.id])
+
+                    val simEntry = if (regionId != null) {
```
la3lma, line 814: This is complex behaviour, it should be put into a separate method or function with a descriptive name. Barring that, a comment describing what's going on should be added.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+                        logger.warn("SimProfile not linked to any region. iccId: {}", simProfile.iccId)
+                        SimEntry(
+                                iccId = simProfile.iccId,
+                                status = NOT_READY,
+                                eSimActivationCode = "Dummy eSIM",
+                                msisdnList = emptyList()
+                        )
+                    }
+
+                    org.ostelco.prime.model.SimProfile(
```
la3lma, line 839: Why can't the original simProfile instance be used? This is just a copy? If it's just a copy, why not use a .copy() method?

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+                                            status = NOT_READY,
+                                            eSimActivationCode = "Dummy eSIM",
+                                            msisdnList = emptyList()
+                                    )
+                                },
+                                { it }
+                        )
+                    } else {
+                        logger.warn("SimProfile not linked to any region. iccId: {}", simProfile.iccId)
+                        SimEntry(
```
la3lma, line 831: duplicated code, see above.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
-                                    iccId = simProfile.iccId,
-                                    alias = simProfile.alias,
-                                    eSimActivationCode = simEntry.eSimActivationCode,
-                                    status = simEntry.status)
-                        }
-                    }.fix()
+            Either.monad<StoreError>().binding {
+                simProfiles.bind().map { simProfile ->
+
+                    val regionId = (regionCode ?: map[simProfile.id])
```
la3lma, line 812: Which map re we talking about here? Is it  mapper, a data structure, and irrespective of that, what does it map from and to, and why?    Using the same word in multiple interpretations in close proximity is confusing.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -806,26 +806,43 @@ object Neo4jStoreSingleton : GraphStore {
...
+
+                    val simEntry = if (regionId != null) {
+                        simManager.getSimProfile(
+                                hlr = hssNameLookup.getHssName(regionId),
+                                iccId = simProfile.iccId).fold(
+                                { error ->
+                                    logger.warn("SimProfile not found in SIM Manager DB. region: {}, iccId: {}, error: {}", regionId, simProfile.iccId, error)
+                                    SimEntry(
+                                            iccId = simProfile.iccId,
```
la3lma, line 821: This looks like repeated code (the same identical code exists below). Repeated code like this makes the code more verbose than it should.  Can this repetition  be replaced by a pre-computed value or function generating the value on the fly?

la3lma (changes requested):

tools/prime-admin/src/main/kotlin/org/ostelco/tools/prime/admin/actions/Utils.kt
vihangpatil: This "right" side of Either<Right,Left>.
It does not have any type.
The function signature hence uses "Generics" - `fun <L, R> Either<L, R>.print()`

Whatever the type of "right" be, it will convert it to json string with formatting.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: This is `map` on Collection.
After `.bind()`, you get the "right" value of Either, which is `Collection<SimProfile>`.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: All these state fields listed above are internal to SIM Manager.
They all are mapped to simplified status externally.
The status in modified code is that external status and not the internal status.

vihangpatil: Yes, only `provisionstate` of internal states is used and is mapped to external states.
This is done here - https://github.com/ostelco/ostelco-core/blob/develop/sim-administration/simmanager/src/main/kotlin/org/ostelco/simcards/admin/SimManager.kt#L48-L59

vihangpatil: Whether other states should be used in this mapping implementation details of SIM Manager.
External status is what other modules see.
SIM manager may change the mapping logic without affecting other modules.
But this is not in the scope of this PR.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: `.binding {` is moved from the separate line (810 in old code) to one line (809).

General refactoring of the code is not in the scope of this PR.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: Because they are instances of different classes.
https://github.com/ostelco/ostelco-core/blob/develop/model/src/main/kotlin/org/ostelco/prime/model/Entities.kt#L352-L356

https://github.com/ostelco/ostelco-core/blob/develop/neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/model/Model.kt#L29-L35

vihangpatil: The classes are different because we have cases where a single external model class is split into 2 classes in the Graph DB model - one becomes Entity Node and other is Relation.

vihangpatil: In this case, the external class has an extra field compared to the internal Graph DB model class since values for the extra field(s) comes from SIM Manager DB.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: This is `map` on Collection.
After `.bind()`, you get the "right" value of Either, which is `Collection<SimProfile>`.

vihangpatil: It is mapping List of internal SimProfiles to List of external SimProfiles.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
la3lma: In general, the "boyscout principle" https://deviq.com/boy-scout-rule "Leave your code better than you found it" is always active, to some extent. 

However, in this case it is not necessary to invoke it.   If the code being reviewed  is hard to understand, a request to address that is not  "general refactoring",  it's a suggestion for how to  make this particular PR better.   

The fact that the same pattern appears in multiple locations may indicate that some general refactoring may be beneficial, but as you say, this PR is not about general refactoring.

PR #951, opened 2019-11-20
Delete customer in Stripe if deleted from backend

kmmatwork:


acceptance-tests/src/main/kotlin/org/ostelco/at/common/StripePayment.kt
```diff
@@ -83,7 +83,7 @@ object StripePayment {
      * verify that the correspondng 'setDefaultSource' API works as
      * intended.
      */
-    fun getDefaultSourceForCustomer(stripeCustomerId: String) : String {
+    fun getDefaultSourceForCustomer(customerId: String) : String {
```
vihangpatil, line 86: Makes sense.  I renamed it from `customerId` to `stripeCustomerId` when they were different.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -488,6 +489,26 @@ object Neo4jStoreSingleton : GraphStore {
...
+                                        "Removing corresponding payment profile when removing customer $customerId " +
+                                                "failed with error ${it.message} : ${it.description}")
+                                PartiallyNotDeletedError(type = customerEntity.name,
+                                        id = "Failed to remove corresponding payment profile when removing customer $customerId",
+                                        error = it).left()
+                            }
+                        }.bind()
             }.fix()
         }.unsafeRunSync()
                 .ifFailedThenRollback(transaction)
```
vihangpatil, line 513: I do not think we need new Error since we do not delete partially.  If it fails to delete in Stripe, it will rollback for Neo4j too.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -488,6 +489,26 @@ object Neo4jStoreSingleton : GraphStore {
...
+                        }.flatMapLeft {
+                            if (it is org.ostelco.prime.paymentprocessor.core.NotFoundError) {
+                                /* Ignore. Customer has not bought products yet. */
+                                Unit.right()
+                            } else {
+                                logger.error(NOTIFY_OPS_MARKER,
+                                        "Removing corresponding payment profile when removing customer $customerId " +
+                                                "failed with error ${it.message} : ${it.description}")
+                                PartiallyNotDeletedError(type = customerEntity.name,
+                                        id = "Failed to remove corresponding payment profile when removing customer $customerId",
```
vihangpatil: Setting description string as id will result in weird message because we use id like this in Error class - `message = "$type - $id not deleted."`

vihangpatil (approved):
Thanks

PR #954, opened 2019-11-21
Adds Postgres sql script to add timestamps to 'sim_entries' table

kmmatwork:
Work in progress for the time being.


sim-administration/simmanager/schema/add-timestamps.sql
```diff
@@ -0,0 +1,53 @@
...
+returns trigger as $$
+begin
+  new.tsprovisionstate := now();
+  return new;
+end;
+$$ language plpgsql;
+create trigger trigger_provisionstate
+before update on sim_entries
+for each row
+when (old.provisionstate is distinct from new.provisionstate)
```
la3lma: Would a bit of indentation be possible here? (I don't know sql so well), but illustrating the block structure of the for loop could be useful (that would be true for the other for loops too).

sim-administration/simmanager/schema/add-timestamps.sql
```diff
@@ -0,0 +1,53 @@
+--
+-- Add timstamp column for hlrstate
```
la3lma: I really like this approach!  Very elegant.

la3lma (approved):

sim-administration/postgres/init.sql
```diff
@@ -22,6 +29,9 @@ create table sim_entries (id bigserial primary key,
                           pin2 varchar(4),
                           puk1 varchar(8),
                           puk2 varchar(8),
+                          tsHlrState timestamp default now(),
+                          tsSmdpPlusState timestamp default now(),
+                          tsProvisionState timestamp default now(),
```
vihangpatil, line 34: So, these are not the timestamps for every state transition, but instead the timestamp when the last state transition happened for that state.  If that's sufficient for the use-case,  then good! 👍

vihangpatil (approved):

sim-administration/postgres/README.md
```diff
@@ -0,0 +1,20 @@
+# SIM Manager DB schema
```
la3lma, line 1: Please take a look at  the Codacy suggestions for the markdown file.  They are minor, but will probably make the markdown a little stricter.

PR #955, opened 2019-11-21
Feature/user location info

mpeterss:
This is to use the 3GPP-USER-LOCATION-INFO to get the MCC MNC for a user to decide if we should charge for roaming. Previously I was using the 3GPP-SGSN-MCC-MNC value, but it looks like this is not always sent. Some MME / S-GW do not include this AVP.

This will use 3GPP-SGSN-MCC-MNC if available else 3GPP-USER-LOCATION-INFO


diameter-stack/src/test/kotlin/org/ostelco/diameter/parser/UserLocationTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+        assertEquals("130", userLocation?.geographicLocationType)
+        assertEquals("242", userLocation?.mcc)
+        assertEquals("01", userLocation?.mnc)
+    }
+
+    @Test
+    fun UserLocationParserMalaysia() {
+
+        val norwegianUserLocation = "8205f261a8b705f261006a9cd1"
```
cyriaquebrousse: Is this Malaysia?

mpeterss: yeah, Ctl -C Ctl - V

Fixing

cyriaquebrousse (approved):
🎻

PR #956, opened 2019-11-21
Timestamps in graph

vihangpatil:
Also removed logging of Jumio result and MyInfo data.

### DIFF
--- a/neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
+++ b/neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
@@ -6,7 +6,6 @@ import arrow.core.Either.Right
 import arrow.core.EitherOf
 import arrow.core.fix
 import arrow.core.flatMap
-import arrow.core.getOrElse
 import arrow.core.left
 import arrow.core.leftIfNull
 import arrow.core.right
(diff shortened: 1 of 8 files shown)


neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -1794,7 +1810,6 @@ object Neo4jStoreSingleton : GraphStore {
                                         customerId = customer.id,
                                         data = extendedStatus
                                 )
-                                logger.info(NOTIFY_OPS_MARKER, "Jumio verification succeeded for ${customer.contactEmail} Info: $extendedStatus")
```
vihangpatil, line 1797: Removed logging for Jumio success case.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -1813,7 +1828,9 @@ object Neo4jStoreSingleton : GraphStore {
                                 data = extendedStatus
                         )
                     }
-                    logger.info(NOTIFY_OPS_MARKER, "Jumio verification failed for ${customer.contactEmail} Info: $extendedStatus")
+                    if(updatedScanInformation.scanResult?.verificationStatus != "NO_ID_UPLOADED") {
```
vihangpatil, line 1831: Also, for the failed case, not logging for NO_ID_UPLOADED (when Customer cancels KYC).

vihangpatil:
@prasanthu I have removed logging for Jumio success case as requested by Bjørn.  We would continue to log only if we have a strong reason to do so.   Is that okay, or should I revert that back?  I have left comments so that you can easily find the relevant code.

prasanthu (approved):

PR #957, opened 2019-11-22
Mark SIM profile as installed, as reported by App

vihangpatil:


customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/SimProfilesResource.kt
```diff
@@ -68,6 +68,24 @@ class SimProfilesResource(private val regionCode: String, private val dao: Subsc
...
             }.build()
 
+    @PUT
+    @Path("/{iccId}/installed")
+    @Produces(MediaType.APPLICATION_JSON)
+    fun updateSimProfile(@NotNull
+                         @PathParam("iccId")
+                         iccId: String,
+                         @Auth
+                         token: AccessTokenPrincipal?): Response =
```
la3lma, line 78: It is a good idea to check if the incoming parameter is actually an ICCID. Check if it's a number sequence is of the correct length (19-21 characters, only digits (decimal)).

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -898,12 +916,74 @@ object Neo4jStoreSingleton : GraphStore {
                             ?: NotFoundError(type = simProfileEntity.name, id = iccId).left().bind<SimProfile>()
 
                     update { simProfile.copy(alias = alias) }.bind()
-                    AuditLog.info(customerId = customerId, message = "Updated alias of SIM Profile")
+                    AuditLog.info(customerId = customerId, message = "Updated alias of SIM Profile (iccId = $iccId)")
+                    org.ostelco.prime.model.SimProfile(
```
la3lma: Here a sim profile is being mapped to an almost identical sim profile object.  There should be a comment stating clear what kind of mapping this is and why it is necessary.  Also since most of the fields are being copied, it would be clearer if a builder with defaults from the original instance was being used. That would make it very obvious exactly  what fields are being changed.  In this case I believe it is "status" and "eSimActivationCode", and in the code  starting at line 945 it is nothing?

neo4j-store/src/test/kotlin/org/ostelco/prime/storage/graph/Neo4jStoreTest.kt
```diff
@@ -964,15 +975,22 @@ class Neo4jStoreTest {
                                                             SimProfile(
                                                                     iccId = "iccId",
                                                                     eSimActivationCode = "eSimActivationCode",
-                                                                    status = AVAILABLE_FOR_DOWNLOAD))),
+                                                                    status = AVAILABLE_FOR_DOWNLOAD,
+                                                                    requestedOn = simProfileForNo.requestedOn
```
la3lma, line 979: Deep indentation.

prime/infra/dev/prime-customer-api.yaml
```diff
@@ -422,8 +422,32 @@ paths:
...
+      operationId: "markSimProfileAsInstalled"
+      parameters:
+        - name: regionCode
+          in: path
+          description: "Region code"
+          required: true
+          type: string
+        - name: iccId
+          in: path
+          description: "ICC ID"
```
la3lma: It's "ICCID", not "ICC ID"

prime/infra/prod/prime-customer-api.yaml
```diff
@@ -422,8 +422,32 @@ paths:
...
+      operationId: "markSimProfileAsInstalled"
+      parameters:
+        - name: regionCode
+          in: path
+          description: "Region code"
+          required: true
+          type: string
+        - name: iccId
+          in: path
+          description: "ICC ID"
```
la3lma: ditto

la3lma (changes requested):

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/SimProfilesResource.kt
```diff
@@ -68,6 +68,24 @@ class SimProfilesResource(private val regionCode: String, private val dao: Subsc
                         .responseBuilder()
             }.build()
 
+    @PUT
+    @Path("/{iccId}/installed")
+    @Produces(MediaType.APPLICATION_JSON)
+    fun updateSimProfile(@NotNull
+                         @PathParam("iccId")
```
la3lma, line 75: camelcase is not really the right thing to do. It's called "ICCID", so only upper or lower case is ok, camelcase not so much.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -583,7 +583,13 @@ object Neo4jStoreSingleton : GraphStore {
                                                             iccId = simProfile.iccId,
                                                             status = simEntry.status,
                                                             eSimActivationCode = simEntry.eSimActivationCode,
-                                                            alias = simProfile.alias)
+                                                            alias = simProfile.alias,
+                                                            requestedOn = simProfile.requestedOn,
```
la3lma, line 593: Another thing is that this method is long, and is also highly indented. Both of these factors makes it less simple to read.  Using a builder pattern will make the method shorter, which is good, but also taking measures to reduce indentation levels would increase readability.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/SimProfilesResource.kt
vihangpatil: There are cases such as physical SIM cards where we do not have real ICCID.  We then use UUID with some special prefix to indicate that.  So, we cannot use the validation proposed above.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
vihangpatil: Comments are added at top-level instead of every line where it is used.  Also, used different alias name for the class in import since we have 2 set of classes with the same names.

neo4j-store/src/test/kotlin/org/ostelco/prime/storage/graph/Neo4jStoreTest.kt
vihangpatil: This deep indentation is as per code formatting settings of IDE.  This is not due to the deep nesting of control flow statements.

prime/infra/dev/prime-customer-api.yaml
vihangpatil: Renamed description, but not the variable since it "may" bread API compatibility.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/resources/SimProfilesResource.kt
vihangpatil: This will change code in many places.   Will refactor in separate PR.

neo4j-store/src/test/kotlin/org/ostelco/prime/storage/graph/Neo4jStoreTest.kt
la3lma: Then the code settings are permitting code that is not very readable.  I will approve  in a few seconds, but this is not very readable code.

la3lma (approved):

sgoodwin (approved):

PR #960, opened 2019-11-25
Add new sim provisioning API to support endpoint

prasanthu:


vihangpatil:
Looks good.  Prime supports setting a human-readable "alias" to the Sim Profile.  You should have it as part of API as an optional parameter.  If it is too much hassle, you make skip it in UI but it will remain in API if you ever want to use it in UI in the future.

kmmatwork (approved):

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
```diff
@@ -283,6 +283,59 @@ class RegionsTest {
...
+
+            var regionDetailsList: Collection<RegionDetails> = get {
+                path = "/regions"
+                this.email = email
+            }
+
+            assertEquals(2, regionDetailsList.size, "Customer should have 2 regions")
+            var receivedRegion = regionDetailsList.find {
+                it.status == APPROVED
+            }!!
```
la3lma: I don't know if kotlin will  propagate assertNotNull done on a variable as that variable being not null.  If that is true, then it could be used instead of "!!", since it will bring a little more clarity to the code, and also more clarity to an error message, should it ever happen.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/store/SubscriberDAO.kt
```diff
@@ -61,7 +61,7 @@ interface SubscriberDAO {
 
     fun getSimProfiles(identity: Identity, regionCode: String): Either<ApiError, Collection<SimProfile>>
 
-    fun provisionSimProfile(identity: Identity, regionCode: String, profileType: String?): Either<ApiError, SimProfile>
+    fun provisionSimProfile(identity: Identity, regionCode: String, profileType: String?, alias: String): Either<ApiError, SimProfile>
```
la3lma, line 64: Should add  a comment about what happens if profileType == null.

la3lma, line 64: Also, it would be nice if these methods had documentation for their parameters.

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
```diff
@@ -464,3 +468,59 @@ class CustomerResource {
...
+                            @PathParam("id")
+                            id: String,
+                            @QueryParam("regionCode")
+                            regionCode: String,
+                            @QueryParam("profileType")
+                            profileType: String,
+                            @QueryParam("alias")
+                            alias: String): Response =
+            if (token == null) {
+                Response.status(Response.Status.UNAUTHORIZED)
```
la3lma, line 499: I personally would have prefered a  return Response.status .... .build() .  in each of the branches instead of this implicit generation of a builder that is then built at the very end. I think it'a harder to read..

Also, if the direct return was used, the whole else branch could be dropped, decreasing the level of indentation, which in general increases readability, although in this simple case I'm not certain it would have made much difference.

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
```diff
@@ -464,3 +468,59 @@ class CustomerResource {
...
+                logger.info("${token.name} Creating new SIM profile in region $regionCode & profileType $profileType for customerId: $id")
+                provisionSimProfile(
+                        customerId =id,
+                        regionCode = regionCode,
+                        profileType = profileType,
+                        alias = alias)
+                        .responseBuilder(success = Response.Status.CREATED)
+            }.build()
+
+    private fun provisionSimProfile(customerId: String, regionCode: String, profileType: String, alias: String): Either<ApiError, SimProfile> {
```
la3lma, line 510: This code has deep indentation and complex control flow.  In my opinion it should be broken up into multiple locally defined functions that are composed in a simple manner so that it can be read at a glance what the method does. THat's not possible as it is written now.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -747,7 +747,8 @@ object Neo4jStoreSingleton : GraphStore {
     override fun provisionSimProfile(
             identity: ModelIdentity,
             regionCode: String,
-            profileType: String?): Either<StoreError, ModelSimProfile> = writeTransaction {
+            profileType: String?,
+            alias: String): Either<StoreError, ModelSimProfile> = writeTransaction {
         IO {
```
la3lma, line 752: High indentation, complex control flow.   I find this hard to read. Please simplify.

la3lma (commented):
Some comments about readability.   Readability is important, because without it it is very difficult to catch logical issues.

la3lma:
There are a number of issues reported by codacy. Please review these issues and incorporate suggestions if they are deemed to be relevant (by glancing at them at least some of them seemed releveant).

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
prasanthu: With arrow,  this is the simplest I can get.  Breaking this into smaller functions won't help much.

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
prasanthu: Not a pattern I invented, it is used elsewhere in our code. If you are proposing some coding guidelines, let's make that with solid examples of patterns we use.

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
prasanthu: There is no need to prevent a crash in the test. If an exception occurs won't the test fail ?

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
prasanthu: This patch didn't add anything to the complexity. Feel free to suggest alternatives.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/store/SubscriberDAO.kt
prasanthu: I agree, add a task in notion task for that.

acceptance-tests/src/main/kotlin/org/ostelco/at/jersey/Tests.kt
vihangpatil: If you want to avoid `!!` then a light-weight way is to replace it with `?: fail("Received null region")`.

That avoids `receivedRegion` from being `nullable` and all extra code of `assertNotNull` and `?.` in many places.

vihangpatil (approved):
Did not find any bugs with the code.  There will always be some things I would do differently.  But I don't think they should stop a PR.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/store/SubscriberDAOImpl.kt
```diff
@@ -190,9 +190,9 @@ class SubscriberDAOImpl : SubscriberDAO {
         }
     }
 
-    override fun provisionSimProfile(identity: Identity, regionCode: String, profileType: String?): Either<ApiError, SimProfile> {
+    override fun provisionSimProfile(identity: Identity, regionCode: String, profileType: String?, alias: String): Either<ApiError, SimProfile> {
         return try {
```
kmmatwork, line 194: The `return` can be replaced with `Either<ApiError, SimProfile> =`
` try { ...` as it is only one statement in the function. No bit deal though.

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
kmmatwork: Personally I prefer a `build()` at the end and only one `return`. Make the code easier to read.

customer-endpoint/src/main/kotlin/org/ostelco/prime/customer/endpoint/store/SubscriberDAOImpl.kt
prasanthu: I only added the `alias` parameter to the method. As @la3lma pointed out there are lot of code readability things we can do. Let's do that in a separate task after coming up with some coding guidelines.

customer-support-endpoint/src/main/kotlin/org/ostelco/prime/support/resources/HoustonResources.kt
kmmatwork: I think that if we can avoid the `if (token == null) {` part, then that could help. Means that the `token` check should be moved to a filter or something (as suggested by @la3lma earlier). Should be done in a separate PR though and not in this, as the same pattern is used in all external APIs.

prime/infra/dev/prime-houston-api.yaml
```diff
@@ -250,6 +250,42 @@ paths:
           description: "The customerId of the customer"
           required: true
           type: string
+  "/support/simprofile/{id}":
```
vihangpatil, line 253: This way of REST API URL gives a wrong impression that the "id" is Sim Profile ID.
It is different from the way Customer Endpoint API URLs are defined.
Over there, we have used this as a guideline - https://cloud.google.com/apis/design/resources

But this is not introduced in this PR.  It is for almost all of the URLs in support-api swagger spec.
So, I think we should do it in some other PR and let this PR merge.

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
kmmatwork: @la3lma There is an ongoing effort to improve and reduce the size of the `Neo4jStore` code. Part of @vihangpatil "rule engine" plans.

kmmatwork (approved):

PR #963, opened 2019-11-26
Added timestamps in SimProfile in GraphQL spec

vihangpatil:


sgoodwin (approved):
<3

la3lma:
Even though this has been committed, I'm pointing out that the timestamps stored  in the graph database are  generated using a different clock than the one used in the postgres database. Since these timestamps are both used for detecting the same type of issues, and may end up being compared with each others,  I want to point out that this could lead to issues due to clock skew.   If we ever end up relying heavily on this logic (which currently we're not), then I would suggest that we use a single source of truth for all the timestamps, so that they can be reliably compared.

PR #964, opened 2019-11-26
Feature/retry logic in es2plus client

la3lma:
Feature to fix bug/misfeature:  Some times the ES2+ responds slowly, and this patch gives the client a little more time to get it right before signalling a failure.

Also some refactoring/additional documentation (based on hints from Intellij which checks for such things before committing to VCS).


mpeterss:
What is the timeout for requests set to in the HttpClient?

la3lma:
Timeout is set to ten seconds: 
 httpClient:
        timeout: 10000ms

vihangpatil (commented):
The connection timeout issue can be fixed with just a one-line change in the config.
The current value of 10s is set to `timeout`.
`connectionTimeout` is not set at all and hence defaults to `500ms`.
Please try that in a separate PR before this elaborate solution.

Reference:
https://www.dropwizard.io/en/stable/manual/configuration.html#httpclient

la3lma:
Have submitted a new one-line PR with the change suggested by @vihangpatil. If that does the trick, I will rework this PR into a pure refactoring PR (adding documentation etc.), and remove the retry logic.

PR #967, opened 2019-11-28
DSL for operations on Offer

vihangpatil:


neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
```diff
@@ -2810,11 +2810,20 @@ object Neo4jStoreSingleton : GraphStore {
...
     private val segmentEntity = Segment::class.entityType
 
-    private val offerToSegmentRelation = RelationType(OFFERED_TO_SEGMENT, offerEntity, segmentEntity, None::class.java)
-    private val offerToSegmentStore = RelationStore(offerToSegmentRelation)
+    val offerToSegmentRelation = RelationType(
+            relation = OFFERED_TO_SEGMENT,
+            from = offerEntity,
+            to = segmentEntity,
+            dataClass = None::class.java)
+            .also { UniqueRelationStore(it) }
```
prasanthu, line 2818: What is the purpose of `.also` here. According to IntelliJ the type of  `offerToSegmentRelation` is  `RelationType<Offer, None, Segment>`. What happens to the returned `UniqueRelationStore` ?

prasanthu (approved):
Curious about that 1 point other than that LGTM

neo4j-store/src/main/kotlin/org/ostelco/prime/storage/graph/Neo4jStore.kt
prasanthu: I understand now, it is set as the `RelationStore` for the  `RelationType`

prasanthu: If that is the case, why not use `apply` https://kotlinlang.org/docs/reference/scope-functions.html#apply

vihangpatil: Nice question. 💪
Definition of `Relation` creates a normal Relation Store and stores it in a Map.
Instead, if we want Unique Relation Store, then it's instance has to be created somewhere.
That instance will then overwrite the normal Relation Store in the Map.

This is very confusing and I will fix it in the next PR coming today.
This is from the old code where we created a `Relation` and immediately on the next like created normal or unique relation store.
But I did not want to keep a reference of any relation store to promote the use of DSL.

In the next  PR, I will remove all the `.also { UniqueRelationStore(it) }` and instead put an optional flag - `isUnique` with default value as false.  Then use that flag to create an appropriate relation store instead of creating a normal one first for all and overwriting it unique store for few.

Not fixing it in this PR since it an existing design in many other places.

PR #977, opened 2019-12-02
Add smdp-plus emulator to enable seagull test again

mpeterss:


docker-compose.seagull.yaml
```diff
@@ -97,6 +106,25 @@ services:
...
+    ports:
+      - "18080:8080"
+      - "18081:8081"
+
+  postgres:
+    container_name: postgres
+    build:
+      context: sim-administration/postgres
+      dockerfile: Dockerfile
+    tmpfs: "//var/lib/postgresql/data"
```
la3lma: double slash?   (just a nit, but is it significant in any way, if not then it would be clearer to just use a single slash, and if it is significant, add a comment).

la3lma (approved):
one minor nit.

docker-compose.seagull.yaml
mpeterss: fixed

PR #981, opened 2019-12-04
Backup neo4j data

vihangpatil:


tools/neo4j-admin-tools/src/main/kotlin/org/ostelco/tools/migration/Main.kt
```diff
@@ -3,13 +3,16 @@ package org.ostelco.tools.migration
...
 import java.nio.file.Files
 import java.nio.file.Paths
+import java.time.ZoneOffset
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
 
 fun main() {
-    // neo4jExporterToCypherFile()
-    cypherFileToNeo4jImporter()
+    neo4jExporterToCypherFile("prod")
```
prasanthu, line 11: comment out the call to `neo4jExporterToCypherFile`

prasanthu (changes requested):

tools/neo4j-admin-tools/src/main/kotlin/org/ostelco/tools/migration/Main.kt
vihangpatil: On the contrary, I kept this uncommented just to be safe.
If both methods are commented then there is a risk that someone might uncomment and run the `cypherFileToNeo4jImporter()`.

This was me being explicit that use `neo4jExporterToCypherFile()`.

prasanthu (approved):

PR #983, opened 2019-12-06
Adding PDP-Address in protobuf sent to OCS from OCS gw

mpeterss:
The PDP-Address holds the IP address of the UE.

### DIFF
--- a/ocs-grpc-api/src/main/proto/ocs.proto
+++ b/ocs-grpc-api/src/main/proto/ocs.proto
@@ -70,6 +70,17 @@ enum ResultCode {
   DIAMETER_RATING_FAILED = 5031;
 }
 
+// 3GPP-PDP-Type #: 29.061
+enum PdpType {
+  IPV4 = 0;
+  PPP = 1;
+  IPV6 = 2;
+  IPV4V6 = 3;
+  NON_IP = 4;
+  UNSTRUCTURED = 5;
+  ETHERNET = 6;
+}
+
 message RedirectServer {
   RedirectAddressType redirectAddressType = 1;
   string redirectServerAddress = 2;
@@ -102,7 +113,8 @@ message PsInformation {
   string sgsnMccMnc = 2;
   string imsiMccMnc = 3;
   bytes userLocationInfo = 4;
-  string pdpAddress = 5;
+  PdpType pdpType = 5;
+  bytes pdpAddress = 6;
 }
 
 message ServiceInfo {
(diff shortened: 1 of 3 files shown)


ocs-grpc-api/src/main/proto/ocs.proto
```diff
@@ -102,6 +102,7 @@ message PsInformation {
   string sgsnMccMnc = 2;
   string imsiMccMnc = 3;
   bytes userLocationInfo = 4;
+  string pdpAddress = 5;
```
la3lma: What is this? It is encoded as a string, but is it really something else? An IP number or something, if so I don't think a string is an appropriate carrier in Protobuf (octet stream would be better), but if it's a FQDN then of course text is acceptable (this may become clearer to me as I read through the PR).

ocsgw/src/main/java/org/ostelco/ocsgw/converter/ProtobufToDiameterConverter.java
```diff
@@ -162,6 +162,9 @@ private static void addPsInformation(final CreditControlContext context, CreditC
                 if (psInformation.getUserLocationInfo() != null) {
                     psInformationBuilder.setUserLocationInfo(ByteString.copyFrom(psInformation.getUserLocationInfo()));
                 }
+                if (psInformation.getPdpAddress() != null) {
+                    psInformationBuilder.setPdpAddress(psInformation.getPdpAddress().getHostAddress());
```
la3lma: Is this an FQDN or a raw IP address, if it's a raw IP address then it should be encoded somewhat differently. See the code proposed in "gyorde" (https://github.com/ExploratoryEngineering/gyorde/blob/master/protobuf/gyorde.proto) for a suggestion (handles both IPV4 and IPV6).

la3lma (approved):
Please review comments and consider using something other than string if this is a naked IP address, not an FQDN.  Either way add a comment describing what is stored in the Protobuf field.

ocsgw/src/main/java/org/ostelco/ocsgw/converter/ProtobufToDiameterConverter.java
mpeterss: This holds the IP address of the handset ( UE ).

I will follow Specification #: 29.061 then and encode the full information.

ocs-grpc-api/src/main/proto/ocs.proto
mpeterss: IPv4 address

ocsgw/src/main/java/org/ostelco/ocsgw/converter/ProtobufToDiameterConverter.java
mpeterss: Added the full 3GPP-PDP-TYPE to the message. In our case we will always get this as IPv4, but now it is good for the future :)

